PR #31, opened 2015-08-04
New package "keepassx"

netarmy:
keepassx is a password manager that you can save username and password in an encrypted `.kdb` file, and be able to autofill the form, for example in a browser.

I don't know what branch I should open a pull request to, so I choose the branch `dev/new-packages` which seems the right place. (Both ab3/abbs and git/github are new to me.) I'm sorry if this don't meet the standard.


MingcongBai:
I am happy to confirm that this configuration builds, and the format is all fine. It looks beautiful, and builds successfully.

Merging this. And `keepassx` will be in the repository today. Thanks for your pull request.

PR #45, opened 2015-08-07
Rename ksh to ksh93 to avoid confusion and use alternatives for ksh93

butangmucat:


butangmucat:
Closing, because a new one will be opened - added shadowsocks-go

PR #70, opened 2015-08-23
Package request: tcllib

ekd123:
My first trial. It's a super simple package.

tcllib is the standard Tcl library.

...

What should I say here?


MingcongBai:
1.17 may be factored as ${VER:0:4} or ${VER%.*}, no big deal, but it makes your life easier.

MingcongBai:
Bad `$SRCTBL` too, the source has version 1.17 instead of 1.17.0. Please change your `$VER`.

MingcongBai:
Opening the door. Thanks!

PR #75, opened 2015-08-27
Package request: easytag

ekd123:
Easytag is a GTK+-based tag editor for audio files, which is part of the GNOME project.


MingcongBai:
Again, you don't need `REL=0`, but no big deal here.

PR #81, opened 2015-09-02
gnupg: add dep 'npth'

### LINKED ISSUE #77: gnupg missing dependency
npth is missing in the default install of gnupg, which leads to the following failure.

gpg-agent: error while loading shared libraries: libnpth.so.0: cannot open shared object file: No such file or directory

gpg: can't connect to the agent: IPC connect call failed

ekd123:
This should fix #77


MingcongBai:
Merging, thanks.

PR #96, opened 2015-10-10
Import capstone and binwalk

liushuyu:
New packages: capstone and binwalk.


MingcongBai:
Merged. However additional work will be necessary before we actually build this.

liushuyu:
So, what is that?

PR #150, opened 2016-02-23
[zynaddsubfx, lash, mxml, liblo] New

liushuyu:
"I told u these are for Project Full LMMS on AOSC OS" --@liushuyu


liushuyu:
Upcoming: Carla, but later, because Carla is another hell.

PR #245, opened 2016-05-12
[haveged] New package

liushuyu:


MingcongBai:
Okay. Looks good, I will convert parameters of the systemd startup command to a configuration file later though.

PR #255, opened 2016-06-13
flat-plat: import, 3.20.20160404

lmy441900:
w/o GDM themes and Chrome extensions.


MingcongBai:
Would you mind helping my OCD by capitalizing the first letter in `PKGDES`. And no, if it's plainly a theme (a.k.a. a data package), there would be no `PKGDEP` in this case, as the theme does not contain any `.so` like older GTK+ themes.

lmy441900:
> Would you mind helping my OCD by capitalizing the first letter in `PKGDES`. 
The old guideline of AOSC OS packages said so (you told me that). Anyway I'll fix that.

> if it's plainly a theme (a.k.a. a data package), there would be no `PKGDEP` in this case
:ok:

MingcongBai:
> The old guideline of AOSC OS packages said so (you told me that). Anyway I'll fix that.

🤣 Let's just say I was stupid...

MingcongBai:
Looks good, thanks!

PR #387, opened 2016-09-22
musicbox: new, 0.2.3.2 & BeautifulSoup4: fix python3 error

LER0ever:
CLI Version of Netease CloudMusic

### DIFF
--- a/extra-multimedia/musicbox/autobuild/defines
+++ b/extra-multimedia/musicbox/autobuild/defines
@@ -1,0 +1,6 @@
+PKGNAME=musicbox
+PKGSEC=multimedia
+PKGDEP="requests beautifulsoup4 pycrypto future"
+BUILDDEP="setuptools"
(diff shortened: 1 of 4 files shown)


MingcongBai:
Which version of Python is this package using? Could you mark `NOPYTHON2=yes` or `NOPYTHON3=yes`?

LER0ever:
Is it still necessary if the package runs under both python 2 and 3?

MingcongBai:
> Is it still necessary if the package runs under both python 2 and 3?

You see, this is an application, but not Python modules... So, yes, for the sake of rigor, specify one, if you know that it works on Python 3, then I would think you should choose 3 for this.

So add `NOPYTHON2=yes` to the `defines`.

MingcongBai:
Just had a conversation with @LER0ever, seems like our `beautifulsoup4` is not working properly for Py3. Waiting on his next PR to fix this issue first.

PR #418, opened 2016-09-30
grub: fix hard-coded initramfs probing name

lmy441900:
For kernels' localnames are not "-aosc-main", generally kernels in tree
"linux-kernel-extra", grub-mkconfig cannot probe their corresponding
initramfs.

Change "aosc-main" to the corresponding localname can fix this problem.
Also, other localname specifications are not necessary. Remove.


lmy441900:
According to @Icenowy and @bobcao3 , the corresponding initial RAM disks of the kernels in `linux-kernel-extra` tree cannot be correctly probed. This PR should fix the problem.

MingcongBai:
> This PR should fix the problem.

As in "tested"?

lmy441900:
> As in "tested"?

Yes.

MingcongBai:
Cool. Merging.

PR #451, opened 2016-10-11
grub: strip all strings before kernel version number

lmy441900:
Patch 10008 is intended to fix problems with detecting the corresponding
initrd of kernels with local names other than "aosc-main". However, the
regexp parser for "$version" still keep strings containing numeric
characters. For example:

    vmlinuz-aosc-ck5-4.7.6

will have a `$version` of

    ck5-4.7.6

because of the "5" in "ck5", which will not produce a correct pattern
when detecting initramfs:

    initramfs-ck5-4.7.6-aosc-ck5.img

(should be `initramfs-4.7.6-aosc-ck5.img` instead)

This PR fixes this problem by stripping out *all* characters before
the *real* version number.

### DIFF
--- a/extra-admin/grub/autobuild/patches/10009-Strip-all-strings-before-kernel-version-number.patch
+++ b/extra-admin/grub/autobuild/patches/10009-Strip-all-strings-before-kernel-version-number.patch
@@ -1,0 +1,14 @@
+diff --git a/10_linux.orig b/10_linux
+old mode 100644
+new mode 100755
+index b82ecee..e79af9c
+--- a/10_linux.orig
++++ b/10_linux
+@@ -217,7 +217,7 @@ while [ "x$list" != "x" ] ; do
+   dirname=`dirname $linux`
+   rel_dirname=`make_system_path_relative_to_its_root $dirname`
+   localname=`echo $basename | sed -e "s,vmlinuz-,,g" | sed -e "s,-[0-9].*,,g"`
+-  version=`echo $basename | sed -e "s,^[^0-9]*-,,g"`
++  version=`echo $basename | sed -e "s,.*-,,g"`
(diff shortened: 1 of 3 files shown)


MingcongBai:
Could you bump `$REL`? Then it will be good to merge. Thanks.

lmy441900:
Please merge if you think there's no problem @MingcongBai

extra-admin/grub/autobuild/patches/10009-Strip-all-strings-before-kernel-version-number.patch
```diff
@@ -0,0 +1,14 @@
+diff --git a/10_linux.orig b/10_linux
+old mode 100644
+new mode 100755
+index b82ecee..e79af9c
+--- a/10_linux.orig
```
MingcongBai: Actually, there is an issue here. The file edited should be correspondent to the file in the GRUB source file. See the newest 10008 patch I edited just yesterday.

MingcongBai:
Review added.

extra-admin/grub/autobuild/patches/10009-Strip-all-strings-before-kernel-version-number.patch
lmy441900: Ouch...

lmy441900:
Ouch.

MingcongBai:
Done. Thanks.

lmy441900:
Now in [linux-kernel-extra](https://github.com/AOSC-Dev/linux-kernel-extra) the CK subversion is reserved:

`vmlinuz-aosc-ck-4.11.8-ck2`

Not

`vmlinuz-aosc-ck2-4.11.8`

Because CK patchset has its own versioning system, I keep it be appended after the kernel version and set the localname to `aosc-ck`. This patch somewhat breaks the probe.

PR #474, opened 2016-10-21
luarocks: new, 2.4.1

LER0ever:
LuaRocks 2.4.1
A package manager for Lua modules.


extra-lua/luarocks/autobuild/defines
```diff
@@ -0,0 +1,6 @@
+PKGNAME=luarocks
+PKGDES="a package manager for Lua modules"
```
MingcongBai: Would you capitalize the first letter?

PR #612, opened 2017-04-03
gnome-mines: add missing dependency by sophgn

sophgn:


MingcongBai:
Thank you very much sir.

PR #637, opened 2017-05-03
Package add: NCBI Blast

StephDC:
NCBI-BLAST+ is the local version of the NCBI Blast <blast.ncbi.nlm.nih.gov>


MingcongBai:
Well since you're not done yet... Will take care of it in a little bit.

PR #652, opened 2017-06-03
ncbi-blast+: fix

StephDC:
Sorry for not testing it out... Not expecting that prepare was not executed before the building system was determined.


MingcongBai:
That's okay, thanks none the less!

PR #664, opened 2017-06-20
Add bedtools

StephDC:
Great tool that could get me into my bed earlier.


extra-scientific/bedtools/autobuild/defines
```diff
@@ -0,0 +1,4 @@
+PKGNAME=bedtools
+PKGSEC=science
+PKGDES="A powerful tool for genome arithmetic"
+export prefix=/usr
```
MingcongBai: Use `ABMK="prefix=/usr"`, but in fact this is already defined internally. That said, this line could be removed.

extra-scientific/bedtools/autobuild/defines
```diff
@@ -0,0 +1,4 @@
+PKGNAME=bedtools
+PKGSEC=science
+PKGDES="A powerful tool for genome arithmetic"
```
MingcongBai: Where did the dependencies go?

extra-scientific/bedtools/autobuild/defines
StephDC: Tested on ACBS that the prefix was not defined. Maybe due to the differences in capitalization? From the makefile, it takes prefix but not PREFIX.

extra-scientific/bedtools/autobuild/defines
StephDC: Seems like that only the make test requires samtools as described in Gentoo's ebuild. It just makes on my instance so I don't know if there would be any other dependencies...

MingcongBai: > Seems like that only the make test requires samtools as described in Gentoo's ebuild. It just makes on my instance so I don't know if there would be any other dependencies...

Well to say the least, if it's a compiled ELF binary, that would require `glibc` (which we will not neglect) - just run `ldd` and see for yourself.

extra-scientific/bedtools/autobuild/defines
MingcongBai: > Tested on ACBS that the prefix was not defined. Maybe due to the differences in capitalization? From the makefile, it takes prefix but not PREFIX.

Then, this is still not the way to do it. Add a line `MAKE_AFTER="prefix=/usr"`.

MingcongBai:
The tree is quite a mess... While the PR is fine. Next time please try and `git pull --rebase` when upstream branch (in this case `staging`) is updated.

MingcongBai:
The tree is quite a mess... While the PR is fine. Next time please try and `git pull --rebase` when upstream branch (in this case `staging`) is updated.

PR #710, opened 2017-10-03
dnsmasq: update to 2.78

l2dy:
See https://security.googleblog.com/2017/10/behind-masq-yet-more-dns-and-dhcp.html and http://www.openwall.com/lists/oss-security/2017/10/02/5.


MingcongBai:
Thank you very much! We will take over from here on, and updates will be pushed very soon.

l2dy:
Use AOSA-2017-0051.

l2dy:
Use AOSA-2017-0070.

PR #993, opened 2018-02-05
Fix cookie path to be consistent with fortune-mod

StephDC:
`/usr/share/fortune` is defined in the build file of fortune-mod, but here we put the cookies into `/usr/share/fortunes`
The inconsistency would result in "no fortunes found" error when trying to execute fortune-zh.


MingcongBai:
Would you be able to change the merge branch to `bugfix`?

StephDC:
That's not gonna work well... I'll reset and create a new one.

MingcongBai:
No problems.

PR #994, opened 2018-02-05
Fix the inconsistency between fortune-mod and this regarding cookiedir

StephDC:
There it goes, the second trial


MingcongBai:
Thank you very much sir. just another thing, could you bump `REL=` in `spec` so we could make this an appropriate update?

StephDC:
There it goes. Hope I am doing it right...

Absolutely need a checklist to avoid skipping such steps before opening such PR.

MingcongBai:
Perfect. I have marked a checklist in my TODO.

PR #1040, opened 2018-02-27
Add boost to Wesnoth Dep

StephDC:
It happens that Wesnoth require a working boost even in runtime.

It also happens that the wesnoth in the repo was built for boost 1.58 but we now have 1.62
This might suggest a required REL bump with a boost version bump.


MingcongBai:
Thanks.

PR #1054, opened 2018-03-05
Add postinst with patch to allow detection of other terminals

StephDC:
Hey Fluxbox how can you assume that I have either GNome or KDE already installed?!


MingcongBai:
I would encourage that you include more terminal emulators, as we can see, there are more terminal emulators than listed... https://packages.aosc.io/search/?q=terminal

MingcongBai:
Will merge anyways, as this would improve usability.

PR #1057, opened 2018-03-06
Stop icon scanning attempt in postinst

StephDC:
The fluxbox-generate_menu should have nothing to do
with icons, but it just did that... Just need a -in
switch to stop it from doing so.


MingcongBai:
Thx.

PR #1104, opened 2018-03-29
libatomic-ops: update to 7.6.4

ivmai:


ivmai:
Note: libatomic_ops-riscv.patch is removed because it is now supported in the upstream.

MingcongBai:
Thank you. We have updated the package in e4fad505a443132e630444d174c6940d9bb74d7b.

PR #1204, opened 2018-06-01
systemd: fix wrong syntax in FallbackNTP

LionNatsu:
FallbackNTP is a *space-separated* list of NTP server hostnames or IP addresses.
I don't know how to escape whitespace in MESON_AFTER. It seems difficult.
Let's use pool.ntp.org in the meantime.


LionNatsu:
The misconfiguration also triggers this bug: https://github.com/systemd/systemd/issues/9162.

TL;DR *Network Time Synchronization* may block systemd from shutting down your computer.

MingcongBai:
Thank you sir.

PR #1374, opened 2018-09-11
Version Bump for BLAST and SAMTools

StephDC:
Tested. Compiles just fine.

SAMTools has 1.9 released but it would not compile, thus I upgrade it to 1.8 as of now, and try to figure out what the issue is with 1.9.


liushuyu (approved):
LGTM

Test build pending

liushuyu:
Unfortunately, `ncbi-blast` cannot be built using `boost` 1.67.0

PR #1383, opened 2018-09-17
OpenJDK: Update to 11 (new LTS)

liushuyu:
Work in progress PR.

**DO NOT MERGE** unless explicitly told.

Progress

- [x] Make OpenJDK 11 builds
- [ ] Make OpenJDK 11 runs correctly
   - [x] amd64
   - [x] armel
   - [x] aarch64
   - [ ] ppc64
   - [ ] ppc32
- [ ] Make OpenJDK 8u coexists with 11 correctly
- [ ] Modify dependencies to depend on OpenJDK 8 if necessary

---
This change is [Reviewable](https://reviewable.io/reviews/aosc-dev/aosc-os-abbs/1383)


extra-java/openjdk-8/autobuild/build
```diff
@@ -0,0 +1,189 @@
...
+ln -sfv /etc/ssl/certs/java/cacerts "$PKGDIR"/usr/lib/java/lib/security/cacerts
+
+pushd icedtea-web
+build_icedtea_web
+popd
+
+# RPM is very, very confused by this.
+rm -rf "$PKGDIR"/usr/share/man/ja
+
+cat > "$SRCDIR"/autobuild/conffiles << EOF
```
LionNatsu, line 123: What about a separate conffiles?

liushuyu:
I discovered OpenJFX (8) will install files into `/usr/lib/java/` which will cause problems.

To make two OpenJDKs co-exist will require them to reside in __versioned__ folders and use `alternatives` to symlink to `/usr/lib/java`

This will make `/etc/openjdk` folder confusing. So I will be making several configuration folders for each version of OpenJDK.

Since Oracle dropped support for OpenJDK completely and handed control over to Red Hat. Further patches will be monitored from Red Hat Inc. OpenJDK 8 will be supported well into the Year 2023.

MingcongBai:
Any update on this? @liushuyu

liushuyu:
>Any update on this? @liushuyu

I am waiting for `icedtea-web` for OpenJDK 11 although it might never come out. In the meantime, I will try to test on other architectures (like PowerPC) and cycle the build (build itself several times) on these architectures to see the JIT performance and/or interpreter reliability.

Also, do note the backward compatibility of OpenJDK 11 is worse than I once thought, so extensive testing is required when deciding which package could depend on OpenJDK 11 instead of OpenJDK 8.

And last but not least, OpenJFX has switched a way to be installed alongside with OpenJDK 11, and the OpenJFX 8 for OpenJDK 8 will make OpenJDK 11 crash (OpenJDK 11 will somehow find the files [classpath?] and say the plugin mechanism is removed)

MingcongBai:
Merged into Testing-Proposed for the Fall Cycle. Closing.

PR #1433, opened 2018-11-01
Bugfix: Upgrade Ruby

StephDC:
The profile.d would set $PATH to 2.4.0 but we now have 2.5.0

Should be a minor change and have no impact on the ability to build or any abi change.


MingcongBai:
Thanks.

PR #1463, opened 2018-11-11
plasma-browser-integration: new, 5.14.3

RedL0tus:
Wiki page: https://community.kde.org/Plasma/Browser_Integration

Components necessary to integrate browsers into the Plasma Desktop

-----

By the way, I met some problem while using the following $SRCTBL: 
```
SRCTBL="https://download.kde.org/stable/plasma/$VER/plasma-browser-integration-$VER.tar.xz"
```
When ACBS download the file using aria2 from this URL, it will download two files (one meta and the real file) instead of one, and ACBS will try to decompress the meta but not the actual tarball...


MingcongBai:
`ABTYPE=cmake` is unnecessary here.

MingcongBai:
Disregard the comment above... Build does not pass without specifying this option.

PR #1475, opened 2018-11-17
mitmproxy 4.0.4 and its friends and their friends and their friends

RedL0tus:
A TODO here: Use system brotli for extra-python/brotlipy.


liushuyu (approved):
LGTM

PR #1496, opened 2018-11-24
pkg add: fcitx5-gtk

eatradish:
https://gitlab.com/fcitx/fcitx5-gtk

### DIFF
--- a/fcitx5-gtk/autobuild/defines
+++ b/fcitx5-gtk/autobuild/defines
@@ -1,4 +1,4 @@
 PKGNAME=fcitx5-gtk
 PKGDES="gtk im module and glib based dbus client library"
-PKGDEP="fcitx-5 gtk-2 gtk-3"
+PKGDEP="fcitx5"
 BUILDEP="gtk-2 gtk-3 extra-cmake-modules gobject-introspection git"
--- a/fcitx5-gtk/autobuild/defines
+++ b/fcitx5-gtk/autobuild/defines
@@ -1,0 +1,4 @@
+PKGNAME=fcitx5-gtk
+PKGDES="gtk im module and glib based dbus client library"
+PKGDEP="fcitx5 gtk-2 gtk-3"
+BUILDEP="extra-cmake-modules gobject-introspection git"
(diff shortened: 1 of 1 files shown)


fcitx5-gtk/autobuild/defines
```diff
@@ -0,0 +1,4 @@
+PKGNAME=fcitx5-gtk
+PKGDES="gtk im module and glib based dbus client library"
+PKGDEP="fcitx5 gtk-2 gtk-3"
```
liushuyu:
```suggestion
PKGDEP="fcitx-5 gtk-2 gtk-3"
```

Same issue here, and because you named the dependency `fcitx-5` so here you need to reflect that. Otherwise `apt` will complain about couldn't find the package named `fcitx5`

RedL0tus:
Missing spec, and it should belong to extra-i18n(?

eatradish:
> Missing spec, and it should belong to extra-i18n(?

Yep, but I don't know where to put it.

RedL0tus:
> > Missing spec, and it should belong to extra-i18n(?
> 
> Yep, but I don't know where to put it.

Just like what you did to the other packages...

liushuyu (commented):
Okay, you need to remove b0922ae9c14c44769a12e852da6cd02e1ff18ae4

fcitx5-gtk/autobuild/defines
```diff
@@ -0,0 +1,4 @@
+PKGNAME=fcitx5-gtk
+PKGDES="gtk im module and glib based dbus client library"
+PKGDEP="fcitx-5 gtk-2 gtk-3"
```
liushuyu:
```suggestion
PKGDEP="fcitx5"
```

fcitx5-gtk/autobuild/defines
```diff
@@ -0,0 +1,4 @@
...
+BUILDEP="extra-cmake-modules gobject-introspection git"
```
liushuyu:
```suggestion
BUILDEP="gtk-2 gtk-3 extra-cmake-modules gobject-introspection git"
```

liushuyu (changes requested):
Move `gtk-2 gtk-3` to `BUILDDEP`

liushuyu (approved):

PR #1501, opened 2018-11-25
pkg add: fcitx5-rime

eatradish:
https://gitlab.com/fcitx/fcitx5-rime

### DIFF
--- a/extra-i18n/fcitx5-rime/autobuild/defines
+++ b/extra-i18n/fcitx5-rime/autobuild/defines
@@ -1,4 +1,4 @@
 PKGNAME=fcitx5-rime
 PKGDES="Fcitx5 wrapper for librime"
-PKGDEP="fcitx5 boost librime opencc brise"
+PKGDEP="fcitx5 boost librime brise"
 BUILDDEP="extra-cmake-modules git"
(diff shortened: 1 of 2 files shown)


extra-i18n/fcitx5-rime/autobuild/defines
```diff
@@ -0,0 +1,4 @@
+PKGNAME=fcitx5-rime
+PKGDES="Fcitx5 wrapper for librime"
+PKGDEP="fcitx5 boost librime opencc brise"
```
liushuyu: Explain why `opencc` is required.

eatradish: > Explain why `opencc` is required.

`brise` need `opencc`?....ok, actually can remove it...

eatradish: and maybe need `curl` used for download cloud pinyin etc...

liushuyu: >brise need opencc?....ok, actually can remove it...

 No, `brise` does not require `opencc`. It is `librime` whose dependencies include `opencc`.

liushuyu:
```suggestion
PKGDEP="fcitx5 boost librime brise"
```

liushuyu (changes requested):
Remove redundant deps

PR #1525, opened 2018-12-07
baidu-pcsgo: new package

eatradish:


extra-web/baidupcs-go/autobuild/defines
```diff
@@ -0,0 +1,5 @@
+PKGNAME=baidupcs-go
+PKGSEC=net
+PKGDES="A baidupcs unofficial client"
```
MingcongBai: Change this to "An unofficial client for BaiduPCS or Baidu Yun".

MingcongBai (changes requested):

extra-web/baidupcs-go/autobuild/defines
eatradish: > Change this to "An unofficial client for BaiduPCS or Baidu Yun".

ok

PR #1543, opened 2018-12-15
OpenMPI: Version Bump

StephDC:
To whom that has a PPCBE,
Please test if this package still compiles/works in your arch.


MingcongBai:
Hmm, I will hold off on this for now, this will require some more rebuilds. Please, if possible, add `PKGBREAK` in accordance to reverse dependencies and their current versions, see `extra-libs/boost` for a classic example.

MingcongBai:
OpenMPI will be updated as part of the Winter 2020 iteration. Closing.

Issue #1661, opened 2019-02-14
dotnet-{runtime,sdk}: CVE-2019-0657


l2dy:
**CVE IDs:** [CVE-2019-0657](https://portal.msrc.microsoft.com/en-US/security-guidance/advisory/CVE-2019-0657)

**Other security advisory IDs:** [RHSA-2019:0349-01](https://access.redhat.com/errata/RHSA-2019:0349)

**Descriptions:**
* .dotnet: Domain-spoofing attack in System.Uri (CVE-2019-0657)

https://github.com/dotnet/announcements/issues/97

**Architectural progress:**

- [x] AMD64 `amd64`

KexyBiscuit:
Resolved by 20d998a and f798d88, closing.

l2dy:
Use AOSA-2019-0061.

PR #1675, opened 2019-02-26
dotnet-{runtime,sdk}: .NET Core February Update targeting testing

KexyBiscuit:
Related: #1661.


MingcongBai:
Thanks!

MingcongBai:
Oh wait. Could you please re-target this PR to explosive? We do not accept any version upgrade on Testing unless it is to fix current issues (which in 99% of cases does not involve a version change).

PR #1686, opened 2019-03-01
shadowsocks-libev: update to 3.2.4

KexyBiscuit:
## See also

[Release v3.2.4](https://github.com/shadowsocks/shadowsocks-libev/releases/tag/v3.2.4)


KexyBiscuit:
Retargeting.

MingcongBai:
For the record, this is because `shadowsocks-libev` qualifies as an exception to the update cycle. https://wiki.aosc.io/developers/aosc-os/cycle-exceptions

PR #1885, opened 2019-07-08
svxlink: new, 17.12.2

misaka4e21:
https://www.svxlink.org/
Maybe an [EchoLink](http://www.echolink.org/) replacement on GNU/Linux.


MingcongBai:
Please re-target to explosive, thanks!

PR #1964, opened 2019-08-21
netplan: new, 0.98

KexyBiscuit:


extra-network/netplan/spec
```diff
@@ -0,0 +1,4 @@
+VER=0.98
+# SRCTBL="https://github.com/CanonicalLtd/netplan/archive/$VER.tar.gz"
```
MingcongBai, line 2: What is with this?

MingcongBai (changes requested):

extra-network/netplan/spec
KexyBiscuit: At the time I committed that, CanonicalLtd's version FTBFS, so a fork was used instead. In the next commit this will be uncommented.

MingcongBai:
Closing - this is *not* a rejection.

We are switching to a topic-based maintenance system. Please re-create your PR against our `stable` branch, use our new Pull Request template, and work in accordance to our [maintenance guidelines](https://wiki.aosc.io/developer/packaging/topic-based-maintenance-guideline/).

PR #2192, opened 2020-06-11
ca-certs: Create update-ca-bundle to handle ca-bundle

StephDC:
This changes the responsibility of generating the ca-bundle.crt from
packager to user. It would therefore allow the user to add custom root
CA certificates to their machine without being worried that it would be
overwritten once ca-certs updates. It would also allow easier
integration of other root CA packages.


MingcongBai:
Thanks!

PR #2477, opened 2020-11-03
optenvw64: match GCC/Binutils with Core and newest MinGW libraries

Icenowy:
Topic Description
-----------------

As our optenvw64 haven't been updated for long, this topic tries to bring it to newest mingw libraries, and GCC/Binutils versions matched with our Core.

Package(s) Affected
-------------------

- `binutils+w64`
- `gcc+w64`
- `mingw+w64`

Security Update?
----------------

No

Build Order
-----------

The pre-built mingw+w64 is needed for building gcc+w64 on other architectures, because it needs some bootstrap.

Architectural Progress
----------------------

For toolchain:
- [x] AMD64 `amd64`   
- [x] AArch64 `arm64`

For library:

----

For toolchain:
- [x] AMD64 `amd64`   
- [x] AArch64 `arm64`

For library:


extra-optenvw64/gcc+w64/spec
```diff
@@ -1,4 +1,3 @@
-VER=7.3.0
-REL=2
-SRCTBL="https://ftp.gnu.org/gnu/gcc/gcc-$VER/gcc-$VER.tar.xz"
-CHKSUM="sha256::832ca6ae04636adbb430e865a1451adf6979ab44ca1c8374f61fba65645ce15c"
+VER=9.3.1
+GITSRC="https://github.com/gcc-mirror/gcc"
```
MingcongBai, line 2: Please use the new ACBS [spec format](https://wiki.aosc.io/developer/packaging/acbs/spec-format/).

extra-optenvw64/mingw+w64/spec
```diff
@@ -1,4 +1,3 @@
-VER=5.0.3
-REL=2
+VER=8.0.0
 SRCTBL="https://sourceforge.net/projects/mingw-w64/files/mingw-w64/mingw-w64-release/mingw-w64-v$VER.tar.bz2"
-CHKSUM="sha256::2a601db99ef579b9be69c775218ad956a24a09d7dabc9ff6c5bd60da9ccc9cb4"
```
MingcongBai, line 4: Please use the new ACBS [spec format](https://wiki.aosc.io/developer/packaging/acbs/spec-format/).

extra-optenvw64/binutils+w64/spec
```diff
@@ -1,3 +1,3 @@
-VER=2.29.1
-SRCTBL="https://ftp.gnu.org/gnu/binutils/binutils-$VER.tar.gz"
-CHKSUM="sha256::0d9d2bbf71e17903f26a676e7fba7c200e581c84b8f2f43e72d875d0e638771c"
+VER=2.34
+GITSRC="https://sourceware.org/git/binutils-gdb.git"
```
MingcongBai, line 2: Please use the new ACBS [spec format](https://wiki.aosc.io/developer/packaging/acbs/spec-format/).

MingcongBai (changes requested):

MingcongBai:
Should we build this optenv for !amd64?

Icenowy:
Technically there's no problem for building for non-amd64. Although whether it will be useful will be a problem.

MingcongBai:
@Icenowy Then I'd suggest that we don't waste our time on the other ones?

extra-optenvw64/gcc+w64/spec
Icenowy: > Please use the new ACBS [spec format](https://wiki.aosc.io/developer/packaging/acbs/spec-format/).

I cannot find anything outside the spec.

Do you want to switch to SRCS?

Icenowy:
Theortically we will finally have box86+wine, which can run programs build for x86-32 Windows.

MingcongBai:
@Icenowy I guess for AArch64, we still need it for building Wine. As for our secondary architectures - I'm not sure how useful they would be (maybe when Hangover is ready, we can introduce this for `ppc64el`).

Icenowy:
I don't think AArch64 wine is ready for using PE. Even if we will use PE, we need to add ARM targets to MinGW (currently we only build i386/x86_64).

MingcongBai:
Well nevermind then! Removing other architectures from this PR :D.

Icenowy:
I do not suggest removing arm64, as there's no technical barrier.

MingcongBai:
> I do not suggest removing arm64, as there's no technical barrier.

Okay, re-added.

MingcongBai (approved):

PR #2610, opened 2020-12-13
fuse-3: upgrade to 3.10.1, remove /dev/fuse from package

cthbleachbit:
Topic Description
-----------------

At the time of writing, current version 3.9.0-1 of `fuse-3` on arm64 contains `/dev/fuse` within the package, created by `util/install_helper.sh`. Since `/dev/fuse` on AOSC OS is actually handled by udev, dpkg will refuse to remove the device node and the package will fail to install. Some build environment may have more restrictions than others. In the case where `mknod` in the container is forbidden, `mknod` during build will fail without getting caught by autobuild, thus resulting in a package without `/dev/fuse`.

This PR updates `fuse-3` and `fuse-common` to 3.10.1 and uniformly patches away creation of `/dev/fuse` at build time.

Package(s) Affected
-------------------

`fuse-3` and `fuse-common` in the same build group `fuse-3`: 3.10.1

Architectural Progress
----------------------

- [x] AMD64 `amd64`   
- [x] AArch64 `arm64`

- [x] Loongson 3 `loongson3`

----

- [x] AMD64 `amd64`   
- [x] AArch64 `arm64`

- [x] Loongson 3 `loongson3`

### DIFF
--- a/extra-libs/fuse-3/02-fuse-3/patches/do-not-mknod-dev-fuse-on-install.patch
+++ b/extra-libs/fuse-3/02-fuse-3/patches/do-not-mknod-dev-fuse-on-install.patch
@@ -1,0 +1,16 @@
+diff -Naur fuse-3.10.1.orig/util/install_helper.sh fuse-3.10.1.mod/util/install_helper.sh
+--- fuse-3.10.1.orig/util/install_helper.sh	2020-12-13 15:42:07.149454208 -0600
++++ fuse-3.10.1.mod/util/install_helper.sh	2020-12-13 15:41:07.686333784 -0600
+@@ -30,10 +30,12 @@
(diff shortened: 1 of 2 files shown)


MingcongBai (approved):

MingcongBai:
Approved (since it works), but please make sure again if there’s a Meson option to change this behaviour.

cthbleachbit:
There is only a meson switch `useroot` on whether the command line tools should have setuid bit, and when turned off the device node won't be created as well. We cannot turn off the switch as a normal user needs the suid bit to the CLI programs, so the only resonable way I can think of is patching away mknod entirely.

MingcongBai:
@cthbleachbit Yikes. Fine.

PR #2723, opened 2021-01-24
hollywood: new, 1.21-5 

PegionFish:
Topic Description
-----------------
Launch Byobu
Open a random number of splits, random sizes
In each split, run a noisy text app

Rules:
 - Must work as a non-root user
 - Must display information indefinitely (a la "watch", or cmatrix)
   + Can use a while/true + a timeout
 - Must not OOM a system
 - Must not over load a system
 - Must not be too egregious with I/O
 - Must not require outbound internet access

*copied from https://github.com/dustinkirkland/hollywood/blob/master/README *

Package(s) Affected
-------------------
hollywood

Security Update?
----------------

No

Architectural Progress
----------------------

- [ ] AMD64 `amd64`   

    

----

- [ ] AMD64 `amd64`   


extra-misc/hollywood/autobuild/defines
```diff
@@ -0,0 +1,4 @@
+PKGNAME=hollywood
+PKGSEC=extra
```
MingcongBai: `extra` is not a valid section. Please refer to `sets/section` for a list of valid sections. This probably goes into `utils`.

extra-misc/hollywood/autobuild/defines
```diff
@@ -0,0 +1,4 @@
...
+PKGDEP="apg bmon byobu cmatrix coreutils gawk htop man-db mlocate moreutils mplayer openssh speedometer tree util-linux"
```
MingcongBai: I doubt that all of these are actual dependencies, especially `byobu`, `mplayer`, ... I think these are supposed to serve as frontends? Please only include all those required for `hollywood` to run as dependencies, and all others should go to `PKGSUG` as optional dependencies.

extra-misc/hollywood/autobuild/defines
```diff
@@ -0,0 +1,4 @@
...
+PKGDES="Fill your console with Hollywood melodrama technobabble."
```
MingcongBai: Drop the period point at the end for `PKGDES`.

extra-misc/hollywood/spec
```diff
@@ -0,0 +1,3 @@
+VER=1.21
+SRCS="https://launchpad.net/hollywood/trunk/1.21/+download/hollywood_1.21.orig.tar.gz"
```
MingcongBai: Hardcoded version here, please substitute them with `$VER`.

MingcongBai (changes requested):

extra-misc/hollywood/autobuild/defines
PegionFish: I'm not sure where this should go & extra seems to be a good target. WIP

extra-misc/hollywood/autobuild/defines
PegionFish: These packages are required to run as dependencies as the README shows.

PegionFish: WIP

extra-misc/hollywood/spec
PegionFish: fixed

extra-misc/hollywood/autobuild/defines
MingcongBai: Extra is not even a valid section. Try `utils`.

extra-misc/hollywood/autobuild/defines
MingcongBai: I don't know where you got this from, nothing like that is mentioned in the README, unless you meant "sample Apps."

PR #3139, opened 2021-06-05
PolicyKit 0.119

### LINKED ISSUE #3148: polkit: CVE-2021-3560 fixed in 0.119
### CVE IDs:

* CVE-2021-3560: local privilege escalation using `polkit_system_bus_name_get_creds_sync()`

### Other security advisory IDs:

* RedHat Bugzilla ID [1961710](https://bugzilla.redhat.com/show_bug.cgi?id=1961710)
* [USN-4980-1](https://ubuntu.com/security/notices/USN-4980-1)

### Description - from RH bugzilla

The function polkit_system_bus_name_get_creds_sync is used to get the uid and pid of the process requesting the action. It does this by
sending the unique bus name of the requesting process, which is typically something like “:1.96”, to dbus-daemon. These unique names are
assigned and managed by dbus-daemon and cannot be forged, so this is a good way to check the privileges of the requesting process.

The vulnerability happens when the requesting process disconnects from dbus-daemon just before the call to polkit_system_bus_name_get_creds_sync starts. In this scenario, the unique bus name is no longer valid, so dbus-daemon sends back an error reply. This error case is handled in polkit_system_bus_name_get_creds_sync by setting the value of the error parameter, but it still returns TRUE, rather than FALSE. 

This behavior means that all callers of polkit_system_bus_name_get_creds_sync need to carefully check whether an error was set. If the calling
function forgets to check for errors then it will think that the uid of the requesting process is 0 (because the AsyncGetBusNameCredsData struct
is zero initialized). In other words, it will think that the action was requested by a root process, and will therefore allow it. 

Certain calltraces in polkit were found to be vulnerable, and could be used by any local attacker to run Polkit method calls as if there were run by root.

### Patches:

Update to version 0.119

cthbleachbit:
Topic Description
-----------------

This PR updates Policy Kit to 0.119 with the following fixes:

* Ownership `polkitd:root` enforced on directory `/etc/polkit-1/rules.d` at post-installation
* Resolves CVE-2021-3560

Package(s) Affected
-------------------

* `polkit`: 0.119

Security Update?
----------------

Yesss - Issue / AOSA number pending.

Architectural Progress
----------------------

- [x] AMD64 `amd64`   
- [x] AArch64 `arm64`

- [ ] Loongson 3 `loongson3`

---

- [x] AMD64 `amd64`   
- [x] AArch64 `arm64`

- [ ] Loongson 3 `loongson3`

### DIFF
--- a/extra-admin/polkit/autobuild/beyond
+++ b/extra-admin/polkit/autobuild/beyond
@@ -1,0 +1,2 @@
+abinfo "Setting configuration dir owner to polkitd user and permissions"
+chown -c -R 27:root "${PKGDIR}/etc/polkit-1/rules.d"
--- a/extra-admin/polkit/autobuild/beyond
+++ b/extra-admin/polkit/autobuild/beyond
@@ -1,2 +1,2 @@
 abinfo "Setting configuration dir owner to polkitd user and permissions"
-chown -c -R 27:root "${PKGDIR}/etc/polkit-1/rules.d"
+chown -cvR 27:root "${PKGDIR}/etc/polkit-1/rules.d"
--- a/extra-admin/polkit/autobuild/defines
+++ b/extra-admin/polkit/autobuild/defines
@@ -2,10 +2,14 @@ PKGNAME=polkit
 PKGDES="Application development toolkit for controlling system-wide privileges"
 PKGSEC=admin
 PKGDEP="expat systemd glib linux-pam js-78"
-BUILDDEP="gobject-introspection gtk-doc autoconf-archive intltool"
+BUILDDEP="gobject-introspection gtk-doc autoconf-archive intltool libxslt"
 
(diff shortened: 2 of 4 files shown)


extra-admin/polkit/autobuild/beyond
```diff
@@ -0,0 +1,2 @@
+abinfo "Setting configuration dir owner to polkitd user and permissions"
+chown -c -R 27:root "${PKGDIR}/etc/polkit-1/rules.d"
```
MingcongBai: Add verbosity here.

MingcongBai (changes requested):

MingcongBai:
@cthbleachbit Please also create an issue before you merge - and mark your commits with "fix #..."

MingcongBai (approved):

PR #3399, opened 2021-08-30
bzip2: add pkgconfig

### LINKED ISSUE #3368: bzip2: missing pkgconfig
### Bug description

There is no `bzip2.pc` or similar file in `/usr/lib/pkgconfig`, which will let some build systems like meson can not find `bzip2`.

### Steps to reproduce

Just run ``` $ ls -l /usr/lib/ | grep bzip ``` or ``` $ ls -l /usr/lib/pkgconfig | grep bzip```, then see the result. Or we can check [here](https://packages.aosc.io/files/amd64/stable/bzip2/1.0.8-2).

### Update channel(s)

Stable

### Package and version

```bash
Package: bzip2
Status: install ok installed
Section: utils
Installed-Size: 600
Maintainer: Mingcong Bai <jeffbai@aosc.io>
Architecture: amd64
Version: 1.0.8-2
Depends: glibc (>= 1:2.32+git20201130), zlib (>= 1.2.11-3)
Description: A freely available, patent free, high-quality data compressor
X-AOSC-ACBS-Version: 20201112
X-AOSC-Commit: 45d130ae31
X-AOSC-Autobuild3-Version: 1:20201202
```

WhiredPlanck:
Topic Description
-----------------

Add `bzip2.pc` into the `bzip2` package, so some build system like `meson` will be able to find `bzip2`.

Package(s) Affected
-------------------

`bzip2`

Security Update?
----------------

No

Architectural Progress
----------------------

- [x] AMD64 `amd64`   
- [x] AArch64 `arm64`
    

- [x] Loongson 3 `loongson3`

----

- [ ] AMD64 `amd64`   
- [ ] AArch64 `arm64`

- [ ] Loongson 3 `loongson3`

### DIFF
--- a/base-utils/bzip2/autobuild/prepare
+++ b/base-utils/bzip2/autobuild/prepare
@@ -5,20 +5,3 @@ sed -e 's/^CFLAGS=\(.*\)$/CFLAGS=\1 \$(BIGFILES)/g' \
 abinfo 'Replacing CFLAGS in Makefiles with what Autobuild3 defined...'
 sed -e "s|-O2|${CFLAGS}|g" -i "$SRCDIR"/Makefile
 sed -e "s|-O2|${CFLAGS}|g" -i "$SRCDIR"/Makefile-libbz2_so
-
-abinfo 'Generating bzip2.pc with HERE documents (contents refer from Arch Linux)...'
-cat << EOF > "$SRCDIR"/bzip2.pc
-prefix=/usr
-exec_prefix=/usr
-bindir=\${exec_prefix}/bin
-libdir=\${exec_prefix}/lib
-includedir=\${prefix}/include
-
-Name: bzip2
-Description: A file compression library
-Version: $PKGVER
-Libs: -L\${libdir} -lbz2
-Cflags: -I\${includedir}
-EOF
-
-cat "$SRCDIR"/bzip2.pc
--- a/base-utils/bzip2/autobuild/prepare
+++ b/base-utils/bzip2/autobuild/prepare
@@ -6,6 +6,19 @@ abinfo 'Replacing CFLAGS in Makefiles with what Autobuild3 defined...'
 sed -e "s|-O2|${CFLAGS}|g" -i "$SRCDIR"/Makefile
 sed -e "s|-O2|${CFLAGS}|g" -i "$SRCDIR"/Makefile-libbz2_so
 
-abinfo 'Copying bzip.pc to SRCDIR and patching the version in it...'
-cp -v "$SRCDIR"/../bzip2.pc "$SRCDIR"
-sed -e "s|1.0.6|$PKGVER|" -i "$SRCDIR"/bzip2.pc
+abinfo 'Generating bzip2.pc with HERE documents (contents refer from Arch Linux)...'
+cat << EOF > "$SRCDIR"/bzip2.pc
+prefix=/usr
+exec_prefix=/usr
+bindir=\${exec_prefix}/bin
+libdir=\${exec_prefix}/lib
+includedir=\${prefix}/include
+
+Name: bzip2
+Description: A file compression library
+Version: $PKGVER
+Libs: -L\${libdir} -lbz2
+Cflags: -I\${includedir}
+EOF
+
+cat "$SRCDIR"/bzip2.pc
--- a/base-utils/bzip2/autobuild/prepare
+++ b/base-utils/bzip2/autobuild/prepare
@@ -1,7 +1,11 @@
-sed -e 's/^CFLAGS=\(.*\)$/CFLAGS=\1 \$(BIGFILES)/' -i ./Makefile-libbz2_so
+abinfo 'Patching the CFLAGS in Makefile of libbz2.so...'
+sed -e 's/^CFLAGS=\(.*\)$/CFLAGS=\1 \$(BIGFILES)/g' \
+    -i "$SRCDIR"/Makefile-libbz2_so
 
-sed -e "s|-O2|${CFLAGS}|g" -i Makefile
-sed -e "s|-O2|${CFLAGS}|g" -i Makefile-libbz2_so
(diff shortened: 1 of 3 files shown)


WhiredPlanck:
LGTM (amd64)
图片

base-utils/bzip2/autobuild/prepare
```diff
@@ -1,4 +1,11 @@
...
 
-sed -i "s|-O2|${CFLAGS}|g" Makefile
-sed -i "s|-O2|${CFLAGS}|g" Makefile-libbz2_so
+abinfo 'Replacing CFLAGS in Makefiles with what Autobuild3 defined...'
+sed -e "s|-O2|${CFLAGS}|g" -i "$SRCDIR"/Makefile
+sed -e "s|-O2|${CFLAGS}|g" -i "$SRCDIR"/Makefile-libbz2_so
+
+abinfo 'Copying bzip.pc to SRCDIR and patching the version in it...'
+cp -v "$SRCDIR"/../bzip2.pc "$SRCDIR"
+sed -e "s|1.0.6|$PKGVER|" -i "$SRCDIR"/bzip2.pc
```
liushuyu: Please do not do this. This operation is very fragile and will break any time since you downloaded the `pkgconfig` file from the Fedora `rawhide` branch.

My recommendation is to either bundle the file ourselves or use a [HERE document construct](https://tldp.org/LDP/abs/html/here-docs.html) to generate a `pkgconfig` file.

base-utils/bzip2/spec
```diff
@@ -1,5 +1,7 @@
 VER=1.0.8
-REL=2
-SRCS="tbl::https://sourceware.org/pub/bzip2/bzip2-$VER.tar.gz"
-CHKSUMS="sha256::ab5a03176ee106d3f0fa90e381da478ddae405918153cca248e682cd0c4a2269"
+REL=3
+SRCS="tbl::https://sourceware.org/pub/bzip2/bzip2-$VER.tar.gz \
+      file::rename=bzip2.pc::https://src.fedoraproject.org/rpms/bzip2/raw/rawhide/f/bzip2.pc"
```
liushuyu: If you really want to download this file instead of bundling it, please use a "pinned version" URL instead of this.

For example:
```suggestion
      file::rename=bzip2.pc::https://src.fedoraproject.org/rpms/bzip2/raw/5141dcf62cb95a32f4541b3c417c60427d4d34af/f/bzip2.pc"
```

(Please ignore this comment if you decided to go with the bundling route)

liushuyu (changes requested):

base-utils/bzip2/autobuild/prepare
WhiredPlanck: I use HERE document construct to generate the `pkgconfig` file, thanks ~

base-utils/bzip2/autobuild/prepare
```diff
@@ -1,4 +1,24 @@
...
+sed -e 's/^CFLAGS=\(.*\)$/CFLAGS=\1 \$(BIGFILES)/g' \
+    -i "$SRCDIR"/Makefile-libbz2_so
 
-sed -i "s|-O2|${CFLAGS}|g" Makefile
-sed -i "s|-O2|${CFLAGS}|g" Makefile-libbz2_so
+abinfo 'Replacing CFLAGS in Makefiles with what Autobuild3 defined...'
+sed -e "s|-O2|${CFLAGS}|g" -i "$SRCDIR"/Makefile
+sed -e "s|-O2|${CFLAGS}|g" -i "$SRCDIR"/Makefile-libbz2_so
+
+abinfo 'Generating bzip2.pc with HERE documents (contents refer from Arch Linux)...'
```
liushuyu: You don't need to put this many details in the print out:

```suggestion
# Contents adapted from Arch Linux
abinfo 'Generating bzip2.pc ...'
```

base-utils/bzip2/autobuild/prepare
```diff
@@ -1,4 +1,24 @@
...
+includedir=\${prefix}/include
+
+Name: bzip2
+Description: A file compression library
+Version: $PKGVER
+Libs: -L\${libdir} -lbz2
+Cflags: -I\${includedir}
+EOF
+
+cat "$SRCDIR"/bzip2.pc
```
liushuyu: You probably want to clean this up, since usually, you don't need to inspect the contents of the generated file.

```suggestion

```

base-utils/bzip2/autobuild/prepare
```diff
@@ -1,4 +1,24 @@
...
+bindir=\${exec_prefix}/bin
+libdir=\${exec_prefix}/lib
+includedir=\${prefix}/include
+
+Name: bzip2
+Description: A file compression library
+Version: $PKGVER
+Libs: -L\${libdir} -lbz2
+Cflags: -I\${includedir}
+EOF
```
liushuyu: Actually, you probably want to move this hunk to `build` file and directly land the `bzip2.pc` file to `"$PKGDIR"/usr/lib/pkgconfig` (see L20 in `build`).

liushuyu (changes requested):

base-utils/bzip2/autobuild/prepare
WhiredPlanck: So I summarize what I need to do: 
1. It will be better to move the here document to `autobuild/build`.
2. Simplify the `abinfo`.
3. Cancel the inspection that `cat`-ting the generated file again.

Are these right?

liushuyu: That is correct.

However, if you really want to inspect the generated file, you could use `tee`.

WhiredPlanck: > That is correct.
> 
> However, if you really want to inspect the generated file, you could use `tee`.

Thanks!

liushuyu (approved):

MingcongBai (approved):

liushuyu (approved):

PR #3486, opened 2021-10-01
README: unify spelling

KexyBiscuit:


Icenowy (approved):
This is not a packaging change, so just approve it because LGTM.

PR #3493, opened 2021-10-01
workflows: add new workflow for building packages

KexyBiscuit:
### Purpose

Run `acbs-build` in a new buildkit container to build packages immediately from Pull Requests, effectively smoke-testing them.

### Usage

```
/build <package>
```

Where `package` will be directly passed to `acbs-build`.


KexyBiscuit:
/build aosc-aaa
PS: this won't work.

MingcongBai (approved):

PR #3503, opened 2021-10-02
workflows: remove usage of GITHUB_TOKEN

outloudvi:
#3502 introduces the use of `GITHUB_TOKEN` into the workflow (because of `gh`), but the token comes with a write permission to the repository when used by a maintainer, which may be a potential threat since build processes apparently does not require writing to the repository.

This PR aims to replace `gh` and thus remove the usage of `GITHUB_TOKEN`.

[Check the workflow run](https://github.com/outloudvi/aosc-os-abbs/runs/3775119498).

### DIFF
--- a/.github/workflows/build_on_request.yml
+++ b/.github/workflows/build_on_request.yml
@@ -21,26 +21,28 @@ jobs:
             package = regex.exec(context.payload.comment.body);
             if (package == null) throw "No package found in the command!";
             return package[1];
-          result-encoding: string
-
-      - uses: actions/checkout@v2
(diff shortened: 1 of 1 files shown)


outloudvi:
Closing since the threat seems to be well controlled (since `acbs` has no access to the token).

KexyBiscuit:
Per [GitHub Docs](https://docs.github.com/en/actions/advanced-guides/using-github-cli-in-workflows), only the step with GITHUB_TOKEN environment variable set will have access to it, otherwise the token is not exposed.

PR #3506, opened 2021-10-02
workflows: enable building on issue comments

KexyBiscuit:
- Install GitHub CLI and checkout PR only if the comment was from PR; #3503.
- Starts the job only if the comment starts with /build, not contains to avoid accidental invocation.
- Tweaks.

### DIFF
--- a/.github/workflows/build_on_request.yml
+++ b/.github/workflows/build_on_request.yml
@@ -2,11 +2,11 @@ name: Build package on request
 
 on:
   issue_comment:
-    types: [created, edited]
+    types: [created]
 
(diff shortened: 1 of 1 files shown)


MingcongBai (approved):

outloudvi (commented):
Great! Just saying, actually [`${{}}` is omittable in the if clause](https://docs.github.com/en/actions/learn-github-actions/workflow-syntax-for-github-actions#jobsjob_idif). For example:

``` yml
if: startsWith(github.event.comment.body, '/build ')
```

PR #3518, opened 2021-10-03
plymouth: remove dangling symlink

KexyBiscuit:
Remove dangling symlink from plymouth.


KexyBiscuit:
Shouldn't affect any package.

MingcongBai (approved):

PR #3529, opened 2021-10-05
workflows: use buildkit daily container

### LINKED ISSUE #4: Container images are behind of time for months
There're two ways resolving this problem.

- Publish new container tarballs more frequently.
- Update the container while building.

KexyBiscuit:
Add a new workflow to build buildkit containers every day, upload to GitHub Container Registry for consumption.

Supersedes #3525, #3527, closes AOSC-Dev/aosc-os-docker-files#4.

### DIFF
--- a/.github/workflows/build_on_request.yml
+++ b/.github/workflows/build_on_request.yml
@@ -4,6 +4,10 @@ on:
   issue_comment:
     types: [created]
 
+permissions:
+  issues: write
+  pull-requests: write
+
 jobs:
   get-packages:
     if: ${{(github.event.comment.body == '/build' && github.event.issue.pull_request) || startsWith(github.event.comment.body, '/build ')}}
--- a/.github/workflows/build_on_request.yml
+++ b/.github/workflows/build_on_request.yml
@@ -48,7 +48,6 @@ jobs:
                 args += data.toString();
               }
             };
-            options.cwd = process.env.GITHUB_WORKSPACE;
 
             await exec.exec('git', ['diff', '--name-only', 'stable', 'HEAD', '*/spec'], options);
             const regex = /.*\/(.*)\/.*/gm;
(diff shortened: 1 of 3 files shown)


.github/workflows/build_on_request.yml
```diff
@@ -84,7 +83,7 @@ jobs:
   build:
```
outloudvi, line 87: since there is [a `permission` directive](https://docs.github.com/en/actions/learn-github-actions/workflow-syntax-for-github-actions#permissions), we can limit the `GITHUB_TOKEN` for this job to `read` (or possibly even `none` since it's a public repo).

.github/workflows/buildkit-daily.yml
```diff
@@ -0,0 +1,95 @@
+name: Publish buildkit daily container
+
+on:
+  schedule:
+    - cron: '0 17 * * *'
+  workflow_dispatch:
```
outloudvi, line 6: I don't recall clearly but I [personally experienced](https://github.com/outloudvi/mw2fcitx/commit/bfad233133659e5f3e2a6728a9c3746c53893f0b) that setting `workflow_dispatch: null` might cause some problems, like there will be no button on the Actions page to trigger this. Please validate it.

.github/workflows/buildkit-daily.yml
```diff
@@ -0,0 +1,95 @@
...
+
+env:
+  REGISTRY: ghcr.io
+  IMAGE_NAME: ${{github.repository_owner}}/aosc-os-buildkit-daily
+
+jobs:
+  tarball:
+    runs-on: [self-hosted, aosc-os, x64]
```
outloudvi, line 14: I see the runner is tagged `AOSC-OS` instead of `aosc-os`, will it be a problem?

outloudvi (commented):
Not really knowing a lot about custom runners or the AOSC tarball build process, so just leaving some comments inline.

.github/workflows/buildkit-daily.yml
KexyBiscuit: Runners are also tagged as X64 and Linux by default, won't be a problem, have tested.

.github/workflows/buildkit-daily.yml
KexyBiscuit: Leaving it blank has worked well for multiple workflows I own... Docs says so: https://docs.github.com/en/actions/learn-github-actions/events-that-trigger-workflows#example

.github/workflows/build_on_request.yml
KexyBiscuit: Will change the token to [restrictive](https://docs.github.com/en/actions/security-guides/automatic-token-authentication#permissions-for-the-github_token), thanks!

KexyBiscuit: All fixed, will port this to all repositories I manage. Thanks again!

outloudvi:
:lgtm:, looking forward to see this integration!

outloudvi (approved):

PR #3539, opened 2021-10-08
workflows-build: build packages automatically

KexyBiscuit:
Now this workflow works as a check. On every push, it compares the branch with stable and build all modified packages.

### DIFF
--- a/.github/workflows/README.md
+++ b/.github/workflows/README.md
@@ -17,7 +17,7 @@ When replying the command under an issue, packages are built from `stable` branc
 
 If packages are specified, only these packages will be built. If no packages are specified and the command is replied under a pull request, all packages which have spec file changed will be built.
 
-If you want the workflow runs as a status check, add `[ci]` to the beginning of your pull request title.
+If you would like for the workflow to run as a status check, prefix your pull request title with [ci].
 
 ## Examples
 
@@ -58,7 +58,7 @@ If you want the workflow runs as a status check, add `[ci]` to the beginning of 
 
 ## Explanation on the codename
 
-The cake is a lie[^2]! Though Project Cake is an automation joining the ongoing efforts[^3], it will only work in simple scenarios.
+The cake is a lie[^2]! Though Project Cake is an effort to augment AOSC OS's ongoing maintenance automation project[^3], it will only work in simple scenarios.
 
 [^1]: https://github.com/actions/runner/issues/241#issuecomment-556845290
 [^2]: https://en.wikipedia.org/wiki/The_cake_is_a_lie
--- a/.github/workflows/README.md
+++ b/.github/workflows/README.md
@@ -58,6 +58,8 @@ If you want the workflow runs as a status check, add `[ci]` to the beginning of 
 
 ## Explanation on the codename
 
-[The cake is a lie](https://en.wikipedia.org/wiki/The_cake_is_a_lie)! Though Project Cake is an automation joining [the ongoing efforts](https://wiki.aosc.io/developer/automation/), it will only work in simple scenarios.
+The cake is a lie[^2]! Though Project Cake is an automation joining the ongoing efforts[^3], it will only work in simple scenarios.
 
 [^1]: https://github.com/actions/runner/issues/241#issuecomment-556845290
+[^2]: https://en.wikipedia.org/wiki/The_cake_is_a_lie
+[^3]: https://wiki.aosc.io/developer/automation/
--- a/.github/workflows/README.md
+++ b/.github/workflows/README.md
@@ -1,4 +1,4 @@
-# Workflow Build Packages (Project Revive)
+# Workflow Build Packages (codename Project Cake)
 
 ## Syntax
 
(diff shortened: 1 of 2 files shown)


.github/workflows/build-packages.yml
```diff
@@ -1,6 +1,8 @@
-name: Build packages on request
```
MingcongBai, line 1: on demand *

MingcongBai (changes requested):

MingcongBai (dismissed):

.github/workflows/build-packages.yml
```diff
@@ -53,7 +62,7 @@ jobs:
               }
             };
 
-            await exec.exec('git', ['diff', '--name-only', 'stable', 'HEAD', '*/spec'], options);
+            await exec.exec('git', ['diff', '--name-only', 'origin/stable', 'HEAD', '*/spec'], options);
```
outloudvi, line 69: Not on this change, but I see some files like [`core-devel/gcc/02-compiler/overrides/usr/lib/gcc/specs/hardened-cc1`](https://github.com/AOSC-Dev/aosc-os-abbs/blob/stable/core-devel/gcc/02-compiler/overrides/usr/lib/gcc/specs/hardened-cc1), which might also be matched to `*/spec`. Will it be a problem?

.github/workflows/build-packages.yml
```diff
@@ -10,39 +12,46 @@ permissions:
 
 jobs:
   get-packages:
-    if: ${{(github.event.comment.body == '/build' && github.event.issue.pull_request) || startsWith(github.event.comment.body, '/build ')}}
+    if: ${{github.event.pull_request || (github.event.comment.body == '/build' && github.event.issue.pull_request) || startsWith(github.event.comment.body, '/build ')}}
```
outloudvi: I think there is not such context as `github.event.issue` (but only [`github.event.issues`](https://docs.github.com/en/actions/learn-github-actions/events-that-trigger-workflows#issues)), but since this PR does not introduce it first, is it already working?

.github/workflows/build-packages.yml
```diff
@@ -1,6 +1,8 @@
-name: Build packages on request
+name: Build packages
 
 on:
+  pull_request:
```
outloudvi, line 4: Please specify detailed types. Some updates like `assigned`, `unassigned`, `labeled`, `unlabeled` etc do not make sense to rebuild the packages.

.github/workflows/build-packages.yml
KexyBiscuit: It'll send incorrect packages to the pipeline, though not a big problem from my point of view.

.github/workflows/build-packages.yml
KexyBiscuit: It's the issue object in https://docs.github.com/en/developers/webhooks-and-events/webhooks/webhook-events-and-payloads#issue_comment.

.github/workflows/build-packages.yml
KexyBiscuit: > By default, a workflow only runs when a pull_request's activity type is opened, synchronize, or reopened.

https://docs.github.com/en/actions/learn-github-actions/events-that-trigger-workflows#pull_request

outloudvi (commented):
Left some inline comments. Actually I'm a little puzzled that it mixes `on.pull_request` and `on.issue_comment` together and filtered some events on each case... I would suggest it be tested before merged, but debugging it on-the-fly is also okay since it does not seem to create bad side effects.

.github/workflows/build-packages.yml
outloudvi: Oh, I got it, thanks!

.github/workflows/build-packages.yml
KexyBiscuit: `*/spec` won't match that file... It only matches files named exactly `spec`.

.github/workflows/README.md
```diff
@@ -54,4 +56,10 @@ If packages are specified, only these packages will be built. If no packages are
 
 - [ACBS](https://wiki.aosc.io/developer/packaging/acbs/)
 
+## Explanation on the codename
+
+The cake is a lie[^2]! Though Project Cake is an automation joining the ongoing efforts[^3], it will only work in simple scenarios.
```
MingcongBai: Though Project Cake is an effort to augment AOSC OS's ongoing maintenance automation project [^3], it will only work in simple scenarios.

MingcongBai (changes requested):

.github/workflows/README.md
```diff
@@ -17,6 +17,8 @@ When replying the command under an issue, packages are built from `stable` branc
 
 If packages are specified, only these packages will be built. If no packages are specified and the command is replied under a pull request, all packages which have spec file changed will be built.
 
+If you want the workflow runs as a status check, add `[ci]` to the beginning of your pull request title.
```
MingcongBai: If you would like for the workflow to run as a status check, prefix your pull request title with `[ci]` .

MingcongBai (changes requested):

.github/workflows/README.md
KexyBiscuit:
```suggestion
If you would like for the workflow to run as a status check, prefix your pull request title with [ci].
```

.github/workflows/README.md
KexyBiscuit:
```suggestion
The cake is a lie[^2]! Though Project Cake is an effort to augment AOSC OS's ongoing maintenance automation project[^3], it will only work in simple scenarios.
```

MingcongBai (approved):

PR #3544, opened 2021-10-09
workflows-build: fail if nothing to upload

KexyBiscuit:
Assume no file found in `/debs` is an error.


outloudvi:
The code seems okay, but I'm not sure if we really need this - is there a case when the build succeeds but no package is built?

KexyBiscuit:
@outloudvi Autobuild3 sometimes doesn't return non-zero exit code when failed, this is just a workaround.

outloudvi (approved):

PR #3558, opened 2021-10-12
landrop: new, 0.4.0

WhiredPlanck:
Topic Description
-----------------

Introduce `landrop`, a tool allows you to drop files to the devices on your LAN.

Package(s) Affected
-------------------

`landrop`

Security Update?
----------------

No

Test Build(s) Done
------------------

**Primary Architectures**

- [x] AMD64 `amd64`   
- [x] AArch64 `arm64`
    

**Secondary Architectures**

- [x] Loongson 3 `loongson3`

Update(s) Uploaded to Stable
----------------------------

**Primary Architectures**

- [ ] AMD64 `amd64`   
- [ ] AArch64 `arm64`

**Secondary Architectures**

- [ ] Loongson 3 `loongson3`

### DIFF
--- a/extra-network/landrop/autobuild/defines
+++ b/extra-network/landrop/autobuild/defines
@@ -2,5 +2,3 @@ PKGNAME=landrop
 PKGSEC=utils
 PKGDES="A tool allows you to drop files to the devices on your LAN"
 PKGDEP="qt-5 libsodium gcc-runtime"
-
-QTPROJ_AFTER="CONFIG+=force_debug_info"
--- a/extra-network/landrop/autobuild/defines
+++ b/extra-network/landrop/autobuild/defines
@@ -3,4 +3,6 @@ PKGSEC=utils
 PKGDES="A tool allows you to drop files to the devices on your LAN"
 PKGDEP="qt-5 libsodium gcc-runtime"
 
-QTPROJ_AFTER="CONFIG+=force_debug_info"
+QTPROJ_AFTER="CONFIG+=force_debug_info \
+              INCLUDEPATH+=$INCLUDE \
+              LIBS+=-L$LIBDIR -lsodium"
--- a/extra-network/landrop/autobuild/defines
+++ b/extra-network/landrop/autobuild/defines
@@ -1,5 +1,5 @@
 PKGNAME=landrop
-PKGSEC=network
+PKGSEC=utils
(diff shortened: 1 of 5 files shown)


extra-network/landrop/spec
```diff
@@ -0,0 +1,4 @@
+VER=0.4.0
+SRCS="tbl::https://github.com/LANDrop/LANDrop/archive/v{VER}.tar.gz"
```
KexyBiscuit:
```suggestion
SRCS="tbl::https://github.com/LANDrop/LANDrop/archive/v$VER.tar.gz"
```

WhiredPlanck:
LGTM on AMD64

image

WhiredPlanck:
**Note**: If you would like to test this topic, you could install the mobile app from https://landrop.app to your device, and install the package(s) from this topic, then follow the app's instructions.

extra-network/landrop/autobuild/defines
```diff
@@ -0,0 +1,6 @@
+PKGNAME=landrop
+PKGSEC=utils
+PKGDES="A tool allows you to drop files to the devices on your LAN"
+PKGDEP="qt-5 libsodium gcc-runtime"
+
+QTPROJ_AFTER="CONFIG+=force_debug_info"
```
MingcongBai: Let's add this to Autobuild3, and include the Autobuild3 update in this topic.

MingcongBai (changes requested):

extra-network/landrop/autobuild/defines
WhiredPlanck: Great, I will handle it.

WhiredPlanck: P.S. It seems that `qmake PREFIX=/path/to/custom/dir` didn't take effect in this case. I have no idea with it yet.

MingcongBai (approved):

WhiredPlanck:
Add test result of 83f4c05 :

Parameter `CONFIG+=force_debug_info` is necessary especially when `.pro` file explicitly specifies the build type as `release` or `debug_and_release` to make sure that Autobuild3 can build up a debug package.

For example, building `fritzing`, which a package specifies the build type as `debug_and_release`, without the parameter above will fail. Add `CONFIG+=force_debug_info` to `etc/autobuild/ab3_defcfg` can be helpful to handle it.

image

**UPDATE** : Here is the run test result
image

PR #3559, opened 2021-10-12
workflows-tarball: initial workflow

KexyBiscuit:
Generate all recipes of tarballs, except those with NVIDIA drivers, on AMD64, every month.

### DIFF
--- a/.github/workflows/tarball-monthly.yml
+++ b/.github/workflows/tarball-monthly.yml
@@ -1,4 +1,4 @@
-name: Publish tarball monthly
+name: Publish tarballs monthly
 
 on:
   schedule:
@@ -79,7 +79,7 @@ jobs:
           chmod 644 ~/.ssh/config ~/.ssh/config.d/aosc
           chmod 600 ~/.ssh/id_ed25519
 
-      - name: Build and download tarball
+      - name: Build and download tarball from buildbot
         env:
           RECIPE: ${{matrix.recipe}}
           BUILDBOT: ${{matrix.buildbot}}
--- a/.github/workflows/tarball-monthly.yml
+++ b/.github/workflows/tarball-monthly.yml
@@ -86,7 +86,7 @@ jobs:
           ARCH: ${{matrix.architecture}}
           GITHUB_TOKEN: ${{secrets.GITHUB_TOKEN}}
         run: |
-          ssh $BUILDBOT -tt "
+          ssh -tt $BUILDBOT "
             set -ex \
             && mkdir -p /buildroots/githubactions/$(echo -n $GITHUB_TOKEN | sha256sum | head -c 64) \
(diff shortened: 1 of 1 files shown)


.github/workflows/tarball-monthly.yml
```diff
@@ -0,0 +1,110 @@
+name: Publish tarball monthly
```
MingcongBai: tarballs *

KexyBiscuit:
```suggestion
name: Publish tarballs monthly
```

MingcongBai (dismissed):

.github/workflows/tarball-monthly.yml
```diff
@@ -0,0 +1,110 @@
...
+          cat << EOF >> ~/.ssh/config
+          Host *
+            Include config.d/*
+          EOF
+          echo $DEV_SSH_PRIVATE_KEY | base64 -d > ~/.ssh/id_ed25519
+          echo $DEV_SSH_PUBLIC_KEY > ~/.ssh/id_ed25519.pub
+          chmod 644 ~/.ssh/config ~/.ssh/config.d/aosc
+          chmod 600 ~/.ssh/id_ed25519
+
+      - name: Build and download tarball
```
KexyBiscuit:
```suggestion
      - name: Build and download tarball from buildbot
```

.github/workflows/tarball-monthly.yml
```diff
@@ -0,0 +1,110 @@
+name: Publish tarball monthly
+
+on:
+  schedule:
+    - cron: '0 0 1 * *'
```
outloudvi, line 5: It would be better if the tarball generation can be put at a time that few changes happen on ABBS. Would it help if we change the schedule to fit in a UTC+8 midnight?

outloudvi (dismissed):
Seems okay to me. Also, I have some concerns on the storage limit of workflow artifacts. Maybe we can reduce the retention time or clean up old tarballs to fix that.

.github/workflows/tarball-monthly.yml
KexyBiscuit: 00:00 UTC is 08:00 in China Standard Time, 17:00 in Pacific Daylight Time and 01:00 in London, I think it's not an inconvenient time. We have contributors all over the world 😋.

KexyBiscuit:
> Seems okay to me. Also, I have some concerns on the storage limit of workflow artifacts. Maybe we can reduce the retention time or clean up old tarballs to fix that.

Storage is unlimited for free, all artifacts are stored for 90 days, so there'll be at most 3 months' tarballs.

outloudvi (approved):
Thanks for your considerate answers. By the way, does this PR need a `workflow` label?

KexyBiscuit:
> Thanks for your considerate answers. By the way, does this PR need a `workflow` label?

This PR was from the time which doesn't have that... Added.

PR #3586, opened 2021-10-18
workflows-build: allow only one run WIP per PR

KexyBiscuit:
https://docs.github.com/en/actions/learn-github-actions/workflow-syntax-for-github-actions#example-using-concurrency-to-cancel-any-in-progress-job-or-run-1


outloudvi (commented):
Generally seems okay. One question: is there any reason to put it on `get-packages`, not the whole workflow or other job steps?

KexyBiscuit:
@outloudvi Good question, I should put it on the whole workflow.

outloudvi (approved):
Not seeing further concerns for now.

PR #3589, opened 2021-10-18
workflows-lgtm: only check packages

KexyBiscuit:
Non-package contributions needs no lgtm label.


outloudvi (dismissed):

outloudvi (approved):
The logic makes sense so I'm giving lgtm.

What I want to append is that, since `context` is available in `actions/github-script`, I would recommend getting rid of `${{}}` and also get it better formatted (e.g. add line breaks properly).

PR #3590, opened 2021-10-18
workflows-lgtm: move if logic outside script

KexyBiscuit:


KexyBiscuit:
Smoke test passed.

outloudvi (approved):
It's okay if smoke tests is passing. But as said by [GitHub docs](https://docs.github.com/en/actions/learn-github-actions/expressions):

> When you use expressions in an if conditional, you may omit the expression syntax (${{ }})

Well, I would really like to see it even easier to read over a style fix, actually...

KexyBiscuit:
@outloudvi Removing `${{}}` results in a corrupted YAML file. GitHub tolerates that, but Visual Studio Code doesn't.

outloudvi:
> @outloudvi Removing `${{}}` results in a corrupted YAML file. GitHub tolerate that, but Visual Studio Code doesn't.

Oh heck... 😅

PR #3592, opened 2021-10-18
workflows-build: support workflow dispatch

KexyBiscuit:
It really hurts debugging GitHub Actions.

Anyway. This update adds workflow dispatch support. User could manually or programmatically dispatch the workflow with package names as input. As always, package names are spitted by whitespace ` `.


KexyBiscuit:
Test passed on my org.

MingcongBai (approved):

PR #3593, opened 2021-10-18
workflows-build: compare PR with common parent

KexyBiscuit:
... instead of current HEAD of stable, builds only packages relevant to the PR.


KexyBiscuit:
Test passed on my org.

MingcongBai (approved):

PR #3768, opened 2022-01-24
glib: disable FAM support since 2.70.2

WhiredPlanck:
Topic Description
-----------------

This pull request is to disable GLib's FAM (File system monitoring) support.

The original reason for this change is that the current FAM support in GLib breaks Firefox's WebExtensions threads with new new GLibC (version 2.34) and make various addons cannot work properly.

We discussed for this, and then found that Fedora has disabled GLib's FAM support since version 2.67, explaining as follows:

> FAM support in GLib requires Gamin component, which has various issues
>     and is unmaintained for years. GLib is the only component that requires
>     it in RHEL. So let's do not build the optional glib2-fam subpackage on
>     RHEL so Gamin can be dropped out from there. I am not doing the same
>     change for Fedora as there are more components that require Gamin. So
>     the subpackage can be held there yet for some time. The only reason why
>     this change hasn't been done earlier is the fact that FAM support was
>     needed for DConf functionality when the home directory is on NFS. But
>     this shouldn't be a problem nowadays thanks to the following commit:
>     https://gitlab.gnome.org/GNOME/glib/-/commit/7427bb71.

So finally according to the result of Community's discussions, we decide to disable FAM support in GLib since 2.70.2 to make the world better.

In addition, this PR will make following adjustments to GLib:
- Disable debugging for [its code is not suitable for a production](https://bugzilla.redhat.com/show_bug.cgi?id=1936339)
- Disable SELinux support for all (distribution policy)
- Enable sysprof (capture) support for Stable (package requirement)
- Build all default libraries (GIO modules) for Stable (distribution policy)

Package(s) Affected
-------------------

```
glib
```

Security Update?
----------------

No

Test Build(s) Done
------------------

**Primary Architectures**

- [x] AMD64 `amd64`   
- [x] AArch64 `arm64`
    

**Secondary Architectures**

- [x] Loongson 3 `loongson3`

Update(s) Uploaded to Stable
----------------------------

**Primary Architectures**

- [ ] AMD64 `amd64`   
- [ ] AArch64 `arm64`

**Secondary Architectures**

- [ ] Loongson 3 `loongson3`


WhiredPlanck:
LGTM on AMD64

MingcongBai (approved):

PR #3825, opened 2022-02-13
livekit-x11-base: new, 1

eatradish:
Topic Description
-----------------

livekit-x11-base for AOSC OS Livekit Desktop

Package(s) Affected
-------------------

livekit-x11-base: 1

Security Update?
----------------

No

Test Build(s) Done
------------------

**Primary Architectures**

Update(s) Uploaded to Stable
----------------------------

**Primary Architectures**

- [ ] Architecture-independent `noarch`


Icenowy (commented):
I don't think it necessary to create a new base meta package.

In addition the wallpaper may be changed.

eatradish:
> I don't think it necessary to create a new base meta package.
> 
> In addition the wallpaper may be changed.

The problem is that some of the `retro-x11-base` configuration files need to be rewritten in the script, which will make the livekit-generated script complicated, so I think we should build this package

Icenowy:
I don't think a lot of configuration files are needed to be updated.

eatradish:
> I don't think a lot of configuration files are needed to be updated.

If you don't use `retro-x11-base` and add related packages directly to livekit.lst, the configuration file needs to be written to the livekit generation script, and if you use the `retro-x11-base` package, some configurations (such as desktop wallpaper) need to be changed again.

Icenowy:
> > I don't think a lot of configuration files are needed to be updated.
> 
> If you don't use `retro-x11-base` and add related packages directly to livekit.lst, the configuration file needs to be written to the livekit generation script, and if you use the retro-x11-base package, some configurations (such as desktop wallpaper) need to be changed again.

I tried to use retro-x11-base in recipe, and the things that need to be changed are quite few,

PR #4010, opened 2022-06-01
Devel Drop Python 2 Survey 20220531

WhiredPlanck:
Topic Description
-----------------

This PR is to drop Python 2 dependency for packages in `extra-devel` category as much as possible.

In addition, `sandsifter` and `gdb` still depend on Python 2. For `sandsifter`, this project is too old and it's inappropriate to migrate to Python 3, since it can only generate scripts in Python 2 format, so this PR just fixes build with GLibC 2.34. For `gdb`, according to the investigation and as @Icenowy said, it is not time to switch to Python 3 yet, so this PR doesn't hack it.

Package(s) Affected
-------------------

- `capstone` v4.0.2-4
- `codelite` v16.0.0+1-1
- `gringo` v5.4.1-2
- `keystone`v0.9.2-4
-  `ngs` v2.10.8-2
- `z3` v4.8.10-2
- `sandsifter` v1.04-2

Security Update?
----------------

No

Build Order
-----------

```
extra-devel/capstone
extra-devel/sandsifter
```

Test Build(s) Done
------------------

**Primary Architectures**

- [x] AMD64 `amd64`   
- [x] AArch64 `arm64`
    

**Secondary Architectures**

- [ ] Loongson 3 `loongson3`
- [ ] RISC-V 64-bit `riscv64`

Update(s) Uploaded to Stable
----------------------------

**Primary Architectures**

- [ ] AMD64 `amd64`   
- [ ] AArch64 `arm64`

**Secondary Architectures**

- [ ] Loongson 3 `loongson3`
- [ ] RISC-V 64-bit `riscv64`


MingcongBai:
Awaiting your test results.

extra-devel/capstone/autobuild/beyond
```diff
@@ -0,0 +1,5 @@
+abinfo "Making capstone Python 3 bindings ..."
+cd "$SRCDIR"/bindings/python
```
MingcongBai, line 2: Consider using functions exported in `build/15-python.sh`.

extra-devel/capstone/autobuild/beyond
```diff
@@ -0,0 +1,5 @@
...
+python3 setup.py install -O1 --root="$PKGDIR" --prefix=/usr
+
+cd "$SRCDIR"
```
MingcongBai: Not necessary.

extra-devel/sandsifter/autobuild/patches/0001-fix-build-with-glibc-2.34.patch
```diff
@@ -0,0 +1,22 @@
+diff --git a/injector.c b/injector.c
```
MingcongBai, line 1: Whose patch is this? Follow our commit convention:

```
sandsifter: (whose patch?) ...

- What did it fix and how?

Ref: ...
```

MingcongBai (changes requested):

extra-devel/sandsifter/autobuild/patches/0001-fix-build-with-glibc-2.34.patch
WhiredPlanck: This is the original patch. The co-author helps me create this patch.

WhiredPlanck:
Codelite LGTM on AMD64

图片

extra-devel/capstone/autobuild/beyond
WhiredPlanck: Python 3 binding here can be considered as a standalone source tree, I think it may be the reason why build outside the tree with absolute path will end up with failure.

MingcongBai (approved):

PR #4159, opened 2022-08-27
Linux kernel and tools 5.19.4 - fix crash during kvm boot on some intel CPUs

cthbleachbit:
Topic Description
-----------------

This PR updates main linux kernel and tools to 5.19.4 and addressed following issues:

- Introduced in 5.18, `X86_KERNEL_IBT` is a kernel security feature that hardens control flow integrity for 11th and 12th gen Intel processors. Prior to 5.19.4 and 6.0~rc2, objtool incorrectly replaced an `ENDBR` instruction in `arch/x86/kvm/emulate.c` in `fastop()` with NOP, causing KVM failures on aforementioned platforms. This problem is specific to Intel processors that supports control-flow integrity protection. ~~If you have appended `ibt=off` or `ibt=warn` to your kernel boot parameters, it's now safe to re-enable IBT by default.~~ IBT is now off by default. Out of tree modules that ship object code in binary form cannot be recompiled to support IBT and will cause problem. Notably nvidia will break whenever IBT is enabled on an affected processor.
- For some reason, kernel `perf` tool was not correctly linked against python 3 support due to python3 building process appending `-fpermissive` (which is obviosly not valid in C and causes GCC to emit an warning) to CFLAGS and kernel appending `-Werror` (which turned the said warning into a fatal error). `-Werror` has been disabled from this point for `perf` to workaround this issue,

Package(s) Affected
-------------------

linux-kernel kernel-tools linux+kernel

Build Order
-----------

linux-kernel kernel-tools linux+kernel

Test Build(s) Done
------------------

**Primary Architectures**

- [x] AMD64 `amd64`   
- [x] AArch64 `arm64`

**Secondary Architectures**

- [x] Loongson 3 `loongson3`
- [x] RISC-V 64-bit `riscv64`

Update(s) Uploaded to Stable
----------------------------

**Primary Architectures**

- [x] AMD64 `amd64`   
- [x] AArch64 `arm64`

**Secondary Architectures**

- [x] Loongson 3 `loongson3`
- [x] RISC-V 64-bit `riscv64`


cthbleachbit:
### Reference: Upstream changes merged in 5.19.4:

- https://lore.kernel.org/lkml/20220823080128.867380224@linuxfoundation.org/
- https://lore.kernel.org/lkml/20220823080128.907850538@linuxfoundation.org/

### Reference: Other sightings
- https://bbs.archlinux.org/viewtopic.php?id=276699

MingcongBai (approved):

PR #4397, opened 2023-01-31
v2raya: new, 2.0.1

### LINKED ISSUE #4369: pakreq: v2rayA
URL: https://github.com/v2rayA/v2rayA
https://v2raya.org/

Description: An easy-to-use and powerful v2Ray client

RedL0tus:
Topic Description
-----------------

Introduce v2raya, a Project V client with WebUI (closes #4369).

Package(s) Affected
-------------------

`v2raya: new, 2.0.1`

Security Update?
----------------

No

Test Build(s) Done
------------------

**Primary Architectures**

- [x] AMD64 `amd64`   
- [x] AArch64 `arm64`
    

**Secondary Architectures**

- [x] Loongson 3 `loongson3`
- [x] RISC-V 64-bit `riscv64`

Update(s) Uploaded to Stable
----------------------------

**Primary Architectures**

- [x] AMD64 `amd64`   
- [x] AArch64 `arm64`

**Secondary Architectures**

- [x] Loongson 3 `loongson3`
- [x] RISC-V 64-bit `riscv64`


MingcongBai:
`v2raya.service` LGTM on AArch64:

图片

`v2raya-lite.service` (WTF is this?) hangs at GeoIP data download with no network activity:

图片

CalciteW:
> `v2raya.service` LGTM on AArch64:
> 
> 图片
> 
> `v2raya-lite.service` (WTF is this?) hangs at GeoIP data download with no network activity:
> 
> 图片

Problems with `v2raya-lite.service` can be solved by `systemctl restart`

CalciteW:
this problem with `v2raya-lite.service` can be solved by running `systemctl restart`

MingcongBai:
> this problem with `v2raya-lite.service` can be solved by running `systemctl restart`

@RedL0tus I think it could still be worth reporting this issue to upstream (since you mentioned that the service was poorly written). Another option is to just forgo the `-lite` service.

RedL0tus:
> > this problem with `v2raya-lite.service` can be solved by running `systemctl restart`
> 
> @RedL0tus I think it could still be worth reporting this issue to upstream (since you mentioned that the service was poorly written). Another option is to just forgo the `-lite` service.

Well actually it's my bad. `v2ray-lite.service` is supposed to be an user service, it works just fine in that case. But I still think it requires some modification:
 - I found that when launching `v2raya` with `--lite`, it defaults to `${HOME}/.local/share/v2ray/`. Maybe we should explicitly set that to something like `${HOME}/.config/v2raya`?
 - Its `After` field is a bit weird. The "lite" mode shouldn't have any routing-related capability, yet `v2raya-lite.service` uses the same values for `v2raya.service`.

MingcongBai:
@RedL0tus To the second point, I'm not sure what you mean - are you talking about the launch path?

To the final point, please consult the upstream for possible clarification/fixes.

RedL0tus:
> @RedL0tus To the second point, I'm not sure what you mean - are you talking about the launch path?
> 
> To the final point, please consult the upstream for possible clarification/fixes.

@MingcongBai  LMAO I deleted the most important part for the second point by mistake.

I mean its configuration path is not explicitly stated in the unit file. The default one uses `v2ray` instead of `v2raya` as directory name. IMO that might not be a good idea.

MingcongBai:
@RedL0tus I see. Would the $HOME variable work in systemd units? That might be something to double check.

RedL0tus:
> @RedL0tus I see. Would the $HOME variable work in systemd units? That might be something to double check.

@MingcongBai It should work, or we could use the [`%E` specifier](https://www.freedesktop.org/software/systemd/man/systemd.unit.html#Specifiers) here (and that might be a better idea).

MingcongBai:
@RedL0tus Got it.

extra-network/v2raya/autobuild/defines
```diff
@@ -0,0 +1,7 @@
+PKGNAME=v2raya
+PKGSEC=net
+PKGDES="A client of Project V with web GUI"
```
MingcongBai: A client for

extra-network/v2raya/spec
```diff
@@ -0,0 +1,4 @@
+VER=2.0.1
+SRCS="https://github.com/v2rayA/v2rayA/archive/refs/tags/v${VER}.tar.gz"
+CHKSUMS="sha256::5f79a42c4a720dec778a5c03c925dbbeb8d0fedebbb38ddf90d2d9712480b6a4"
+CHKUPDATE="git::url=https://github.com/v2rayA/v2rayA.git;pattern=v2.*"
```
MingcongBai: Use anitya whenever possible. They are the most reliable.

MingcongBai (changes requested):

extra-network/v2raya/autobuild/defines
RedL0tus: Done

MingcongBai (approved):

PR #4693, opened 2023-08-29
feishin: add desktop entry and icon

OriginCode:
Topic Description
-----------------

Fix desktop entry and add an icon.

Package(s) Affected
-------------------

`feishin` 0.3.0-1

Security Update?
----------------

No

Test Build(s) Done
------------------

**Primary Architectures**

- [x] AMD64 `amd64`   
- [x] AArch64 `arm64`
    

Update(s) Uploaded to Stable
----------------------------

**Primary Architectures**

- [ ] AMD64 `amd64`   
- [ ] AArch64 `arm64`

### DIFF
--- a/app-multimedia/feishin/autobuild/overrides/usr/share/applications/feishin.desktop
+++ b/app-multimedia/feishin/autobuild/overrides/usr/share/applications/feishin.desktop
@@ -1,0 +1,10 @@
+[Desktop Entry]
+Type=Application
+Name=Feishin
+Comment=A music player client for self-hosted streaming services
+Comment[zh_CN]=自建音乐串流服务客户端
+Comment[zh_TW]=自架音樂串流服務用戶端
+Exec=/usr/bin/feishin
+Icon=feishin
+Terminal=false
+Categories=AudioVideo;Audio;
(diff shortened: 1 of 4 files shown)


app-multimedia/feishin/autobuild/overrides/usr/share/applications/feishin.desktop
```diff
@@ -0,0 +1,10 @@
+[Desktop Entry]
+Type=Application
+Name=Feishin
+Comment=A music player client for self-hosted streaming services
+Comment[zh_CN]=自建音乐串流服务客户端
```
MingcongBai, line 5: Feishin fails to detect mpv that is already installed on the system. Yes, the upstream already noted this, but it seems silly for a user to do things that the distro should have already handled (setting the mpv path before the first launch).

Upstream statement:

"Upon startup you will be greeted with a prompt to select the path to your MPV binary. If you do not have MPV installed, you can download it [here](https://mpv.io/installation/) or install it using any package manager supported by your OS. After inputting the path, restart the app."

What the user sees:

```
MPV REQUIRED

Set your MPV executable location below and restart the application.
```

Yes, the user could have went ahead and found the upstream repo, and read their README. But why would we assume for anyone to have thought of doing so? We packaged Feishin using our *system dependency*, mpv, and it should be detected right away - otherwise, why should the dependency be there in the first place?

As it stands, I suggest leaving this package in this topic (I have removed the previous packages from `stable`); another way to handle this would be to patch in a `localSetting` item, `mpv_path`, per [/src/renderer/features/action-required/components/mpv-required.tsx:10](https://github.com/jeffvli/feishin/blob/1acfa93f1addb004de3734c2c1486279c2bec716/src/renderer/features/action-required/components/mpv-required.tsx#L10C9-L10C24).

OriginCode:
LGTM on amd64, successfully used `/usr/bin/mpv` as default MPV executable path and skipped the action required page.
圖片

MingcongBai:
LGTM on AArch64.

MingcongBai (approved):

PR #4914, opened 2023-12-13
mimalloc: new package

felixonmars:
Topic Description
-----------------

mimalloc: new package

Package(s) Affected
-------------------

mimalloc

Security Update?
----------------

No

Test Build(s) Done
------------------

**Primary Architectures**

- [x] AMD64 `amd64`   
- [x] AArch64 `arm64`
    

**Secondary Architectures**

- [ ] Loongson 3 `loongson3`
- [ ] RISC-V 64-bit `riscv64`

### DIFF
--- a/runtime-common/mimalloc/autobuild/defines
+++ b/runtime-common/mimalloc/autobuild/defines
@@ -1,0 +1,9 @@
+PKGNAME=mimalloc
+PKGDES="A high-performance general purpose allocator"
+PKGSEC=libs
+PKGDEP="glibc"
+
+# Default behavior (-DMI_INSTALL_TOPLEVEL=OFF) puts libraries inside a 
+# sub-directory in /usr/lib, which will cause unnecessary rebuilds on update.
+# Use -DMI_INSTALL_TOPLEVEL=ON to install libraries in /usr/lib.
+CMAKE_AFTER="-DMI_INSTALL_TOPLEVEL=ON"
--- a/runtime-common/mimalloc/spec
+++ b/runtime-common/mimalloc/spec
@@ -1,0 +1,4 @@
+VER=2.1.2
+SRCS="git::commit=tags/$VER::https://github.com/microsoft/mimalloc"
+CHKSUMS="SKIP"
+CHKUPDATE="anitya::id=233872"


app-devel/mimalloc/spec
```diff
@@ -0,0 +1,4 @@
+VER=2.1.2
+SRCS="tbl::https://github.com/microsoft/mimalloc/archive/refs/tags/v$VER.tar.gz"
```
MingcongBai: Please use Git source here - GitHub's automatically generated tarballs are not reliable.

app-devel/mimalloc/autobuild/defines
```diff
@@ -0,0 +1,6 @@
+PKGNAME=mimalloc
+PKGDES="A compact general purpose allocator with excellent performance"
```
MingcongBai: A high-performance general purpose allocator.

Avoid fluff/advertisement from the upstream.

app-devel/mimalloc/autobuild/defines
```diff
@@ -0,0 +1,6 @@
...
+PKGSEC=devel
```
MingcongBai: `libs` is probably better since this will be a runtime package.

app-devel/mimalloc/autobuild/defines
MingcongBai, line 1: Move to `runtime-common` since this is a runtime package.

app-devel/mimalloc/autobuild/defines
```diff
@@ -0,0 +1,6 @@
...
+PKGDEP="glibc"
+
+CMAKE_AFTER="-DMI_INSTALL_TOPLEVEL=ON"
```
MingcongBai: Please add a comment explaining that this will be default upstream behaviour (upstream marks the "OFF" setting as deprecated).

MingcongBai (changes requested):

runtime-common/mimalloc/autobuild/defines
```diff
@@ -0,0 +1,7 @@
+PKGNAME=mimalloc
+PKGDES="A high-performance general purpose allocator"
+PKGSEC=libs
+PKGDEP="glibc"
+
+# Default behavior to put libs inside a subdir will cause unnecessary rebuilds on update
```
MingcongBai: Default behavior (-DMI_INSTALL_TOPLEVEL=OFF) puts libraries inside a sub-directory in /usr/lib, which will cause unnecessary rebuilds on update. Use -DMI_INSTALL_TOPLEVEL=ON to install libraries in /usr/lib.

MingcongBai (changes requested):

MingcongBai (approved):

PR #4789, opened 2023-10-24
Linux Kernel 6.6

MingcongBai:
Topic Description
-----------------

This topic updates Linux Kernel to 6.6, including support for `loongarch64`.

This topic is part of the effort to merge #4701.

- [x] Kernel update to 6.6
- [x] Kernel tools update to 6.6
- [x] Fix Linux firmware and AMD/Intel microcode loading
- [x] NVIDIA tested? #4808
- [x] VirtualBox tested?
- ~Incorporate Linux Surface patch set?~

Package(s) Affected
-------------------

- `kernel-build-common`
- `linux-kernel`
- `linux+kernel`
- `linux-firmware`
- `kernel-tools`

Security Update?
----------------

No

Build Order
-----------

```
#buildit kernel-build-common linux-kernel linux+kernel linux-firmware kernel-tools
```

Test Build(s) Done
------------------

**Primary Architectures**

- [x] AMD64 `amd64`   
- [x] AArch64 `arm64`

**Secondary Architectures**

Architectural progress for "secondary," or experimental ports does not impede on merging of this topic.

- [x] MIPS R6 64-bit (Little Endian) `mips64r6el`
- [x] PowerPC 64-bit (Little Endian) `ppc64el`
- [x] RISC-V 64-bit `riscv64`

PR #5050, opened 2023-12-30
Kmod revert to 30

Cyanoxygen:
Topic Description
-----------------

This PR reverts `kmod` back to version 30, due to a fatal regression which might renders the system unbootable.

Package(s) Affected
-------------------

- `kmod` 1:30

Security Update?
----------------

No

Test Build(s) Done
------------------

**Primary Architectures**

- [x] AMD64 `amd64`
- [ ] AArch64 `arm64`

**Secondary Architectures**

- [ ] Loongson 3 `loongson3`
- [ ] RISC-V 64-bit `riscv64`

### DIFF
--- a/app-admin/kmod/autobuild/defines
+++ b/app-admin/kmod/autobuild/defines
@@ -19,3 +19,9 @@ AUTOTOOLS_AFTER="--disable-experimental \
                  --with-xz \
                  --with-zlib \
                  --with-openssl"
+
+# Reverted 4774b31b8a58ce65895d586c489c248733b02965 due to a fatal regression
+# discovered which might render the system unbootable
+PKGEPOCH=1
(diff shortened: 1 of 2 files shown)


MingcongBai:
This issue will be addressed from the kernel side, our current mainline (6.4.11) and longterm (6.1.45) kernel versions are affected by a bug where the xz compression uses crc64 and a dictionary size of 2MiB, which is not supported by kmod 31's [in-kernel xz decompressor](https://lwn.net/Articles/946017/) implementation.

- This issue has been addressed with a branch change on 6.4; no fix is made available for 6.1.
- Kernel 6.6 was not affected by this issue.

We will update the kernel to 6.6 (expediting #4789) which sets the default module compression method to zstd and change longterm kernel's module compression method to zstd.

Closing.

PR #5469, opened 2024-03-09
ucx: new, 1.15.0

leavelet:
Topic Description
-----------------
- ucx: new, 1.15.0

Package(s) Affected
-------------------

- ucx: 1.15.0

Security Update?
----------------

No

Test Build(s) Done
------------------

**Primary Architectures**

- [x] AMD64 `amd64`
- [ ] AArch64 `arm64`

**Secondary Architectures**

- [ ] Loongson 3 `loongson3`
- [ ] RISC-V 64-bit `riscv64`

Architectural progress for experimental ports does not impede on merging of this topic.

### DIFF
--- a/runtime-scientific/ucx/autobuild/defines
+++ b/runtime-scientific/ucx/autobuild/defines
@@ -1,0 +1,25 @@
+PKGNAME=ucx
+PKGSEC=libs
+PKGDEP="rdma-core fuse-3"
+BUILDDEP__AMD64="${BUILDDEP} cuda valgrind"
+BUILDDEP__ARM64="${BUILDDEP} cuda valgrind"
+PKGDES="A communication framework for high-bandwidth and low-latency networks"
+PKGSUG="cuda"
+
+AUTOTOOLS_AFTER="--prefix=/usr \
+		 --enable-mt \
+		 --enable-cma \
+  		 --disable-params-check \
+		 --with-rdmacm \
+		 --with-verbs \
+		 --with-mlx5-dv \
+		 --with-rc \
+		 --with-ud \
+		 --with-dc \
+		 --without-go \
+		 --without-java"
+AUTOTOOLS_AFTER__AMD64="${AUTOTOOLS_AFTER} --with-valgrind --with-cuda=/usr/lib/cuda"
+AUTOTOOLS_AFTER__ARM64="${AUTOTOOLS_AFTER} --with-valgrind --with-cuda=/usr/lib/cuda"
+
+# FIXME: Upstream support is currently limited to the following architectures.
+FAIL_ARCH="!(amd64|arm64|ppc64el)"
--- a/runtime-scientific/ucx/spec
+++ b/runtime-scientific/ucx/spec
@@ -1,0 +1,4 @@
+VER=1.15.0
+SRCS="git::commit=tags/v$VER::https://github.com/openucx/ucx"
+CHKSUMS="SKIP"
+CHKUPDATE="anitya::id=16895"


app-network/ucx/autobuild/defines
```diff
@@ -0,0 +1,24 @@
...
+PKGSEC=science
+PKGDEP="gcc glibc libtool pkg-config rdma-core fuse-3"
+BUILDDEP="autoconf automake m4"
+BUILDDEP__AMD64="${BUILDDEP} cuda valgrind"
+BUILDDEP__ARM64="${BUILDDEP} cuda valgrind"
+PKGDES="Unified Communication Library"
+PKGSUG="cuda"
+
+NOSTATIC=0
+AUTOTOOLS_AFTER="--libdir=/usr/lib/ucx \
```
jiegec: Why `/usr/lib/ucx`? I think using the default `/usr/lib` is okay, and we can remote the extra ld.conf.

leavelet: I referred to OpenMPI's packaging scripts (`app-scientific/openmpi`) and I once thought this was the convention of AOSC.

jiegec: OpenMPI is different: you may have multiple MPI implementations, but ucx is unique.

jiegec:
After this pr, we can enable ucx support in openmpi.

jiegec (dismissed):
ucx_perftest works

app-scientific/openmpi/autobuild/defines
```diff
@@ -24,3 +24,7 @@ AUTOTOOLS_AFTER="--sysconfdir=/etc/openmpi \
 AUTOTOOLS_AFTER__RISCV64="${AUTOTOOLS_AFTER} --disable-memchecker --without-valgrind"
 PKGBREAK="arpack-ng<=3.2.0-1 boost<=1.58.0-3 supertuxkart<=0.9.2 \
           valgrind<=3.12.0"
+
+# Upstream support is limited to the following archs
+# otherwise FTBFS with "Unsupported architecture"
+FAIL_ARCH="!(amd64|arm64|ppc64el)"
```
leavelet: RV64 supported will be available in v1.16.0

app-network/ucx/autobuild/defines
```diff
@@ -0,0 +1,24 @@
+PKGNAME=ucx
+PKGSEC=science
+PKGDEP="gcc glibc libtool pkg-config rdma-core fuse-3"
+BUILDDEP="autoconf automake m4"
+BUILDDEP__AMD64="${BUILDDEP} cuda valgrind"
+BUILDDEP__ARM64="${BUILDDEP} cuda valgrind"
+PKGDES="Unified Communication Library"
```
MingcongBai: Best to be a little more descriptive, what does it do? What does it contain?

app-network/ucx/autobuild/defines
```diff
@@ -0,0 +1,24 @@
+PKGNAME=ucx
+PKGSEC=science
+PKGDEP="gcc glibc libtool pkg-config rdma-core fuse-3"
+BUILDDEP="autoconf automake m4"
```
MingcongBai: No need to specify these as they are part of BuildKit.

app-network/ucx/autobuild/defines
```diff
@@ -0,0 +1,24 @@
+PKGNAME=ucx
+PKGSEC=science
+PKGDEP="gcc glibc libtool pkg-config rdma-core fuse-3"
```
MingcongBai: Why is `pkg-config` listed as a dependency? Same goes for `libtool`, does it depend on `libltdl.so.7`?

app-network/ucx/autobuild/defines
```diff
@@ -0,0 +1,24 @@
+PKGNAME=ucx
+PKGSEC=science
```
MingcongBai: Probably a `libs`?

app-network/ucx/autobuild/prepare
```diff
@@ -0,0 +1,2 @@
+abinfo "Running autogen.sh ..."
```
MingcongBai: This is done automatically in Autobuild (unless you specify `RECONF=0`).

app-scientific/openmpi/autobuild/defines
```diff
@@ -24,3 +24,7 @@ AUTOTOOLS_AFTER="--sysconfdir=/etc/openmpi \
 AUTOTOOLS_AFTER__RISCV64="${AUTOTOOLS_AFTER} --disable-memchecker --without-valgrind"
 PKGBREAK="arpack-ng<=3.2.0-1 boost<=1.58.0-3 supertuxkart<=0.9.2 \
           valgrind<=3.12.0"
+
+# Upstream support is limited to the following archs
```
MingcongBai: Rewrite as:

```
# FIXME: Upstream support is currently limited to the following architectures.
```

MingcongBai (changes requested):

app-network/ucx/autobuild/defines
leavelet: It's a unified library for communication, I can't find description for it.

- 'Unified' means it can support various hardware from different vendors with unified abstraction.
- 'Communication' means it is related to communication.
- 'Library' means other apps can use it as a library.

app-network/ucx/autobuild/defines
leavelet: I read this [article](https://wiki.aosc.io/developer/packaging/package-styling-manual/)

> Additional Notes
> When a package has a sole dependency on the GCC Runtime (gcc-runtime) or the GNU C Library (glibc), these dependencies should be included in $PKGDEP.

Maybe I misunderstood it.

app-network/ucx/autobuild/defines
leavelet: No. `pkg-config` was used to deal with path of dynamic libraries. But since @jiegec modified the logic, it is no longer needed.

app-network/ucx/autobuild/prepare
leavelet: Again. I referred to OpenMPI's packaging scripts (app-scientific/openmpi) and I once thought this was the convention of AOSC.

app-network/ucx/autobuild/defines
leavelet: The package is placed under `app-network` by mistake. I'll adjust its location.

leavelet:
Modified according to these conversations

runtime-scientific/ucx/autobuild/prepare
```diff
@@ -0,0 +1 @@
+sh "$SRCDIR"/autogen.sh
```
MingcongBai: Again, drop this file as this is handled automatically by Autobuild.

runtime-scientific/ucx/autobuild/defines
```diff
@@ -0,0 +1,26 @@
+PKGNAME=ucx
+PKGSEC=libs
+PKGDEP="rdma-core fuse-3"
+BUILDDEP__AMD64="${BUILDDEP} cuda valgrind"
+BUILDDEP__ARM64="${BUILDDEP} cuda valgrind"
+PKGDES="A communication framework for high-bandwidth and low-latency networks"
+PKGSUG="cuda"
```
MingcongBai, line 7: What happens if CUDA is not installed? Does the user get an error?

runtime-scientific/ucx/autobuild/defines
```diff
@@ -0,0 +1,26 @@
...
+
+NOSTATIC=0
```
MingcongBai: Why?

runtime-scientific/ucx/spec
```diff
@@ -0,0 +1,4 @@
+VER=1.15.0
+SRCS="git::commit=tags/v$VER::https://github.com/openucx/ucx"
+CHKSUMS=""
```
MingcongBai: Mark `CHKSUMS="SKIP"`.

MingcongBai (changes requested):

runtime-scientific/ucx/autobuild/defines
leavelet: No. CUDA will not be used unless it is explicitly invoked.

runtime-scientific/ucx/autobuild/prepare
leavelet: I thought it was an AOSC convention. Sorry about it.

runtime-scientific/ucx/autobuild/defines
leavelet: Don't know why. I copied that from openmpi spec

MingcongBai (approved):

PR #5687, opened 2024-04-06
adaptivecpp: 24.02.0, new

ouankou:
Topic Description
-----------------

adaptivecpp: 24.02.0, new

Package(s) Affected
-------------------

adaptivecpp: 24.02.0

Security Update?
----------------

No

Build Order
-----------

```
#buildit adaptivecpp
```

Test Build(s) Done
------------------

**Primary Architectures**

- [x] AMD64 `amd64`
- [x] AArch64 `arm64`

**Secondary Architectures**

- [x] Loongson 3 `loongson3`
- [x] RISC-V 64-bit `riscv64`

Architectural progress for experimental ports does not impede on merging of this topic.

- [x] MIPS R6 64-bit (Little Endian) `mips64r6el`

### DIFF
--- a/app-devel/adaptivecpp/autobuild/defines
+++ b/app-devel/adaptivecpp/autobuild/defines
@@ -1,0 +1,17 @@
+PKGNAME=adaptivecpp
+PKGSEC=devel
+PKGDEP="boost gcc-runtime glibc llvm python-3"
+PKGDES="Compiler for C++-based heterogeneous programming models"
+
+BUILDDEP="boost llvm python-3"
+ABTYPE=cmakeninja
+CMAKE_AFTER=" \
+    -DACPP_CONFIG_FILE_GLOBAL_INSTALLATION=ON \
+    -DDEFAULT_TARGETS='omp.accelerated' \
+    -DWITH_ACCELERATED_CPU=ON \
+    -DWITH_CUDA_BACKEND=OFF \
+    -DWITH_LEVEL_ZERO_BACKEND=OFF \
+    -DWITH_OPENCL_BACKEND=OFF \
+    -DWITH_ROCM_BACKEND=OFF \
+    -DWITH_SSCP_COMPILER=OFF \
+"
--- a/app-devel/adaptivecpp/spec
+++ b/app-devel/adaptivecpp/spec
@@ -1,0 +1,4 @@
+VER=24.02.0
+SRCS="git::commit=tags/v${VER}::https://github.com/AdaptiveCpp/AdaptiveCpp.git"
+CHKSUMS="SKIP"
+CHKUPDATE="anitya::id=176075"


jiegec:
Can build and run the example from https://github.com/AdaptiveCpp/AdaptiveCpp/blob/develop/doc/examples.md properly.

app-devel/adaptivecpp/autobuild/defines
```diff
@@ -0,0 +1,17 @@
+PKGNAME=adaptivecpp
+PKGSEC=devel
+PKGDEP="boost gcc-runtime glibc llvm python-3"
+PKGDES="Independent, community-driven compiler for C++-based heterogeneous programming models"
```
MingcongBai: Compiler for C++-based heterogeneous programming models

app-devel/adaptivecpp/autobuild/defines
```diff
@@ -0,0 +1,17 @@
...
+ABTYPE=cmake
```
MingcongBai: Try cmakeninja?

MingcongBai: I would also keep it near CMAKE_AFTER.

app-devel/adaptivecpp/autobuild/defines
```diff
@@ -0,0 +1,17 @@
...
+BUILDDEP="boost llvm python-3"
+
+CMAKE_AFTER=" \
+    -DACPP_CONFIG_FILE_GLOBAL_INSTALLATION=ON \
+    -DDEFAULT_TARGETS='omp.accelerated' \
+    -DWITH_ACCELERATED_CPU=ON \
+    -DWITH_CUDA_BACKEND=OFF \
+    -DWITH_LEVEL_ZERO_BACKEND=OFF \
+    -DWITH_OPENCL_BACKEND=OFF \
```
MingcongBai, line 14: Why disable all these features?

MingcongBai (changes requested):

app-devel/adaptivecpp/autobuild/defines
ouankou: Similar to LLVM's OpenMP GPU offloading capabilities, all these features require the availability of corresponding components (such as CUDA, ROCM, and OpenCL implementations) both at compile time and runtime. However, we cannot guarantee that these components will always be available or that they will match between the compile and runtime environments. The only component consistently available is LLVM OpenMP on the host. Users have to build other features from source on their own machines.

Additionally, SSCP operates on a JIT compilation flow, allowing a single binary to potentially function across multiple backends like OpenCL and CUDA without recompilation. However, we can only support one host backend due to the constraints, and no backend switching would occur. Therefore, it is better to utilize AOT compilation over JIT, which has more overhead.

app-devel/adaptivecpp/autobuild/defines
ouankou: Fixed.

app-devel/adaptivecpp/autobuild/defines
ouankou: Fixed. `cmakeninja` works fine.

app-devel/adaptivecpp/autobuild/defines
MingcongBai: Fair enough.

MingcongBai (approved):

PR #6576, opened 2024-06-11
grub (GRand Unified Bootloader): improve compliance with UEFI Specification

MingcongBai:
Topic Description
-----------------

- grub: use capitalised path in grub-mkrescue
    Per UEFI Specification, section 3.4.1.1:
    ... If FilePathList[0] points to a device that supports the
    EFI_SIMPLE_FILE_SYSTEM_PROTOCOL, then the system firmware will attempt to boot
    from a removable media FilePathList[0] by adding a default file name in the
    form \EFI\BOOT\BOOT{machine type short-name}.EFI.
    While FAT < 32 filesystems are not case sensitive (which grub-mkrescue creates
    as a FAT12 image via mformat with a size of 2.88MiB), it seems that
    some of Loongson's LoongArch-based firmware (namely those found on their
    latest XA61200 boards) seems to treat this file system as case-sensitive. In
    this case, at least the XA61200 board seems incapable of identifying
    /efi/boot/bootloongarch64.efi as a valid removable EFI image.
    In any case, according to the UEFI Specification, all paths and image
    filenames should be capitalised (with the exception of BOOTx64.EFI, according
    to section 3.5.1.1, for some reason) to stay compliant. The Loongson case is
    only one example of users running into buggy firmware and unbootable GRUB
    ISO images.

Package(s) Affected
-------------------

- grub: 2:2.12+unifont15.1.04-2

Security Update?
----------------

No

Build Order
-----------

```
#buildit grub
```

Test Build(s) Done
------------------

**Primary Architectures**

- [ ] AMD64 `amd64`
- [x] AArch64 `arm64`

**Secondary Architectures**

- [x] RISC-V 64-bit `riscv64`

### DIFF
--- a/app-admin/grub/autobuild/patches/0001-Revert-templates-Properly-disable-the-os-prober-by-d.patch
+++ b/app-admin/grub/autobuild/patches/0001-Revert-templates-Properly-disable-the-os-prober-by-d.patch
@@ -1,7 +1,7 @@
 From 2ec97c359ea87a2152e4076770c9ec8f7cbd225d Mon Sep 17 00:00:00 2001
 From: Javier Martinez Canillas <javierm@redhat.com>
 Date: Fri, 11 Jun 2021 12:10:54 +0200
-Subject: [PATCH 01/21] Revert "templates: Properly disable the os-prober by
+Subject: [PATCH 01/22] Revert "templates: Properly disable the os-prober by
  default"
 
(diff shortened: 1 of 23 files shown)


MingcongBai:
Tested good on LoongArch (produces a compliant, bootable LiveKit image).

Cyanoxygen (approved):

PR #6734, opened 2024-06-20
cmake (CMake): rework postinst

MingcongBai:
Topic Description
-----------------

- cmake: fix postinst during intra-minor upgrades
    The previous change to the postinst breaks when the user attempts to upgrade
    cmake from the same minor series (i.e., 3.29.1 => 3.29.2), as the symlinked
    /usr/share/cmake-x.y never gets removed (since it was already a symlink).
    This revision only tests for the existence of such a matching name and removes
    the symlink (as well as real directories) when found, allowing upgrades to
    commence smoothly.
    Also, create symlinks with relative targets.

Package(s) Affected
-------------------

- cmake: 3.29.6-2

Security Update?
----------------

No

Build Order
-----------

```
#buildit cmake
```

Test Build(s) Done
------------------

**Primary Architectures**

- [x] AMD64 `amd64`
- [x] AArch64 `arm64`

**Secondary Architectures**

- [ ] Loongson 3 `loongson3`
- [ ] RISC-V 64-bit `riscv64`

### DIFF
--- a/app-devel/cmake/autobuild/build
+++ b/app-devel/cmake/autobuild/build
@@ -29,11 +29,13 @@ abinfo "Generating postinst script ..."
 MMVER="${PKGVER%.*}"
 cat << EOF > "$SRCDIR"/autobuild/postinst
 for i in 3.{1..${MMVER#*.}}; do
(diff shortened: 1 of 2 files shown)


MingcongBai:
Testers, please reinstall this version repeatedly.

jiegec (approved):

PR #6899, opened 2024-07-03
openssh (OpenSSH): rework previous systemd changes

MingcongBai:
Topic Description
-----------------

- openssh: rework previous systemd changes
    - Drop sshd.socket.
        - The original configuration file is broken (with Accept=no in pairing
        with sshd.service), as sshd doesn't support systemd-style listen fds
        without downstream patches (and it uses inetd-style activation
        instead). [^1]
        - The sshd.socket unit requires manual syncing of the "Port" setting in
        /etc/ssh/sshd_config (not to mention the lack of multi-port support in
        the original .socket unit. Lest this is done, it's confusing for the
        user as they would have to reach for two configuration files before any
        meaningful debugging can be done for a failing socket-style daemon.
        - Debian/Ubuntu used to ship with sshd-socket-generator for generating
        the .socket unit from /etc/ssh/sshd_config, but since this isn't really
        a commonly used method, let's drop this .socket unit for lack of testing
        and not complicate this matter any further. [^2] [^3]
    - Drop ASCII colour markers in postinst. We barely use colours in our other
    packages and the original implementation missed the \033[0m sequence.
    [^1]: https://salsa.debian.org/ssh-team/openssh/-/blob/master/debian/patches/systemd-socket-activation.patch
    [^2]: For an implementation of the sshd-socket-geneator, see openssh 1:9.6p1-3ubuntu3:
    https://launchpad.net/ubuntu/+archive/primary/+sourcefiles/openssh/1:9.6p1-3ubuntu3/openssh_9.6p1-3ubuntu3.debian.tar.xz
    [^3]: aosc-os-presets-base enables sshd.service by default, the convenience of
    the sshd.socket in our out-of-the-box system is moot.

Package(s) Affected
-------------------

- openssh: 9.8p1-2

Security Update?
----------------

No

Build Order
-----------

```
#buildit openssh
```

Test Build(s) Done
------------------

**Primary Architectures**

- [x] AMD64 `amd64`
- [x] AArch64 `arm64`

**Secondary Architectures**

- [x] Loongson 3 `loongson3`
- [x] RISC-V 64-bit `riscv64`

- [x] MIPS R6 64-bit (Little Endian) `mips64r6el`


app-network/openssh/autobuild/overrides/usr/lib/systemd/system/sshd.service
```diff
@@ -7,7 +7,6 @@ After=sshdgenkeys.service network.target
 Type=notify
```
YHNdnzj: OpenSSH 9.8 supports synchronous reload through `sd_notify` protocol. I.e. `Type=notify-reload` can be used (and `ExecReload=` shall be dropped)

YHNdnzj: https://github.com/openssh/openssh-portable/commit/08f579231cd38a1c657aaa6ddeb8ab57a1fd4f5c

KexyBiscuit (approved):
Upgrade from 9.7p1 works, `systemctl reload sshd` works, oma/apt log looks fine, smoke test passes, LGTM.

PR #7026, opened 2024-07-16
iodine: update to 0.8.0

SkyBird233:
Topic Description
-----------------

- iodine: bump REL for topic Revision Marking Guidelines
- iodine: update to 0.8.0

Package(s) Affected
-------------------

- iodine: 0.8.0-1

Security Update?
----------------

No

Build Order
-----------

```
#buildit iodine
```

Test Build(s) Done
------------------

**Primary Architectures**

- [x] AMD64 `amd64`
- [x] AArch64 `arm64`

**Secondary Architectures**

- [x] Loongson 3 `loongson3`
- [x] RISC-V 64-bit `riscv64`


MingcongBai (changes requested):
Both services failed to launch, seems that the CLI usage has changed:

image

SkyBird233:
> Both services failed to launch, seems that the CLI usage has changed

Services of iodine also require options defined in `/etc/sysconfig/iodine-{client,server}`. Works after defining them.

MingcongBai:
> > Both services failed to launch, seems that the CLI usage has changed
> 
> Services of iodine also require options defined in `/etc/sysconfig/iodine-{client,server}`. Works after defining them.

Then something should be defined before hand... If we can't, we'd need to write a debconf prompt for this package to instruct the user.

SkyBird233:
I believe it's not meaningful to start iodine without the user's configuration, unless it's for local testing, as it necessitates setting up real subdomains. 

For reference, iodine-0.7.0 on AOSC OS, 0.8.0 on Arch Linux and 0.7.0 on Debian require additional configuration before starting iodine and its systemd services.

If prompts are still necessary in this case, could you pleasr provide some detailed suggestions?

MingcongBai:
> I believe it's not meaningful to start iodine without the user's configuration, unless it's for local testing, as it necessitates setting up real subdomains.
> 
> For reference, iodine-0.7.0 on AOSC OS, 0.8.0 on Arch Linux and 0.7.0 on Debian require additional configuration before starting iodine and its systemd services.
> 
> If prompts are still necessary in this case, could you pleasr provide some detailed suggestions?

I think at least a debconf prompt (please reference other packages with `autobuild/templates`) which points the user to relevant documentation and configuration files (what do the users need to change before starting the service?) would be good.

app-network/iodine/autobuild/templates
```diff
@@ -0,0 +1,8 @@
+Template: iodine/sercives_config_info
+Type: note
+Description: Extra config required for iodine services
```
MingcongBai:
```
Description: Extra configuration required for iodine services
  Before you start iodine's client and server services via systemd, please edit the following files to your specific needs:
 .
 /etc/sysconfig/iodine-client: Client configuration.
 /etc/sysconfig/iodine-server: Server configuration.
 .
 For details on how to setup the aforementioned configurations, please see iodine(8).
Description-zh_CN.UTF-8: iodine 服务需要额外设置
 在启动 iodine 客户端或服务端前，您需要预先编辑如下配置文件：
 .
 /etc/sysconfig/iodine-client：客户端配置文件
 /etc/sysconfig/iodine-server：服务端配置文件
 .
 有关上述配置的编写方式，请详见 iodine(7)。
```

app-network/iodine/spec
```diff
@@ -1,4 +1,4 @@
-VER=0.7.0
+VER=0.8.0
 SRCS="https://github.com/yarrick/iodine/archive/refs/tags/v$VER.tar.gz"
```
MingcongBai: Use Git tag.

MingcongBai (changes requested):

MingcongBai (approved):
LGTM on AMD64.

PR #7406, opened 2024-07-30
[DONT MERGE] Revert GCC target/PR 81616 for Afterglow

lunzima:
Topic Description
-----------------

Revert GCC target/PR 81616 for Afterglow so older 32-bit x86 devices might get better compiler optimizations.
This patch only applies to behavior of -mtune=generic. So there should be another patch for Autobuild4.
The previous -mtune=bonnell parameter is obviously not suitable for most devices.

Patch should be okay for upstream GCC version 13.3.x, modifications needed for newer versions of GCC.
Note that if you blindly apply the patch to a later version of GCC source code, the generic_cost struct may be incorrectly patched.

The patch looks good on mingw-w64 and crosstool-NG generated toolchains.

Package(s) Affected
-------------------

core-devel/gcc (cross-toolchain could be affected too?)

Security Update?
----------------

No

IDK but there should be a complete recompilation for BuildKit and all other packages.

Test Build(s) Done
------------------

**Primary Architectures**

- [ ] AMD64 `amd64`
- [ ] AArch64 `arm64`

**Secondary Architectures**

- [ ] Loongson 3 `loongson3`
- [ ] RISC-V 64-bit `riscv64`

### DIFF
--- a/core-devel/gcc/01-runtime/patches/0200-i386-revert-PR-target-81616-for-afterglow.patch.i486
+++ b/core-devel/gcc/01-runtime/patches/0200-i386-revert-PR-target-81616-for-afterglow.patch.i486
@@ -1,0 +1,1568 @@
+diff -ruNd a/gcc/config/i386/athlon.md b/gcc/config/i386/athlon.md
+--- a/gcc/config/i386/athlon.md
++++ b/gcc/config/i386/athlon.md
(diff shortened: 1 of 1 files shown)


MingcongBai:
> IDK but there should be a complete recompilation for BuildKit and all other packages.

Yes, this will be done as part of the reboot.

PR #7471, opened 2024-08-03
opendoas: new, 6.8.2

ziyao233:
Topic Description
-----------------

opendoas: new, 6.8.2

Package(s) Affected
-------------------

- opendoas: 6.8.2

Security Update?
----------------

No

Test Build(s) Done
------------------

**Primary Architectures**

- [x] AMD64 `amd64`
- [ ] AArch64 `arm64`

**Secondary Architectures**

- [ ] Loongson 3 `loongson3`
- [ ] RISC-V 64-bit `riscv64`

### DIFF
--- a/app-admin/opendoas/autobuild/build
+++ b/app-admin/opendoas/autobuild/build
@@ -1,0 +1,9 @@
+abinfo "Configure OpenDoas"
+"$SRCDIR"/configure --prefix=/usr \
+	--with-timestamp
+
+abinfo "Building OpenDoas"
+make
+
+abinfo "Installing OpenDoas"
+make install DESTDIR="$PKGDIR"
--- a/app-admin/opendoas/autobuild/defines
+++ b/app-admin/opendoas/autobuild/defines
@@ -1,0 +1,4 @@
+PKGNAME=opendoas
+PKGSEC=admin
+PKGDEP="glibc linux-pam"
+PKGDES="A tool for delegating authority to users and groups"
--- a/app-admin/opendoas/spec
+++ b/app-admin/opendoas/spec
@@ -1,0 +1,4 @@
+VER=6.8.2
+SRCS="git::commit=tags/v$VER::https://github.com/Duncaen/OpenDoas.git"
+CHKSUMS="SKIP"
+CHKUPDATE="anitya::id=205804"
--- a/app-admin/opendoas/autobuild/conffiles
+++ b/app-admin/opendoas/autobuild/conffiles
@@ -1,0 +1,1 @@
+/etc/pam.d/doas


app-admin/opendoas/spec
xtexx, line 1: `CHKUPDATE="anitya::id=205804"`
pls, ref: <https://release-monitoring.org/project/205804/>

app-admin/opendoas/autobuild/build
```diff
@@ -0,0 +1,9 @@
+abinfo "Configure OpenDoas"
+"$SRCDIR"/configure --prefix=/usr \
```
eatradish, line 2: Use `ABTYPE=autotools` ?

app-admin/opendoas/autobuild/defines
```diff
@@ -0,0 +1,4 @@
+PKGNAME=opendoas
+PKGSEC=admin
+PKGDEP="linux-pam"
```
eatradish:
```
 0x0000000000000001 (NEEDED)             Shared library: [libpam.so.0]
 0x0000000000000001 (NEEDED)             Shared library: [libc.so.6]
```

Add `glibc` to PKGDEP

eatradish (changes requested):

app-admin/opendoas/autobuild/build
ziyao233: OpenDoas is not using autotools (no autotools files like configure.ac) and this would cause errors like "failed to reconfigure"

app-admin/opendoas/spec
```diff
@@ -0,0 +1,3 @@
+VER=6.8.2
+SRCS="tbl::https://github.com/Duncaen/OpenDoas/archive/refs/tags/v$VER.tar.gz"
```
stdmnpkg:
```suggestion
SRCS="git::commit=tags/v$VER::https://github.com/Duncaen/OpenDoas"
```

app-admin/opendoas/spec
```diff
@@ -0,0 +1,3 @@
...
+CHKSUMS="sha256::6da058a0e70b7543bc60624389b0b00b686189ec933828c522bf8b2600495a67"
```
stdmnpkg:
```suggestion
CHKSUMS="SKIP"
```

source code hosting platform may regenerate code archive, causing checksum mismatch.

also add an empty line before end of file

stdmnpkg (changes requested):

app-admin/opendoas/autobuild/build
eatradish: > OpenDoas is not using autotools (no autotools files like configure.ac) and this would cause errors like "failed to reconfigure"

emm... Maybe you should refer to the logic of `/usr/lib/autobuild4/templates/10-autotools.sh` , my insight is that it looks for the existence of the configure.ac/configure/autogen.sh/bootstrap file at the same time and if it exists it proceeds with the

eatradish: emm... Maybe you should refer to the logic of `/usr/lib/autobuild4/templates/10-autotools.sh`, my insight is that it looks for the existence of the configure.ac/configure/autogen.sh/bootstrap file at the same time, and if it exists it proceeds to.

Maybe you can go ahead and add `RECONF=0` to try to continue, if it still doesn't work, then the configure is not the standard autotools configure

At this point, you'll need to add a comment to `autobuild/build` explaining the problem you're encountering

app-admin/opendoas/spec
ziyao233: It this really needed?
`git grep 'archive/$VER' | wc -l` gives me 211.

app-admin/opendoas/spec
ziyao233: for the empty line, I would. though I don't see other packages doing this.

app-admin/opendoas/autobuild/build
ziyao233: tried `RECONF=0` in autobuild/defines, doesnt help.

ziyao233:
Changes from v1:

- Add `CHKUPDATE` property (suggested by xtexChooser)
- Add `glibc` to `PKGDEP` (suggested by eatradish)
- Add empty line for files (suggested by stdmnpkg)

app-admin/opendoas/spec
ziyao233: GitHub and GitLab generate stable tarballs, dont think it is a problem.

stdmnpkg: see [package styling manual chapter 3.2](https://wiki.aosc.io/developer/packaging/package-styling-manual/#source-variables)

eatradish: > GitHub and GitLab generate stable tarballs, dont think it is a problem.

Github seems to recompress the source in a way that you different regions, so the resulting checksum may be incorrect for repeated builds, so you should use the Git source first

However, for opendoas, it's also possible to use a tarball created upstream, such as https://github.com/Duncaen/OpenDoas/releases/download/v6.8.2/opendoas-6.8.2.tar.gz

ziyao233:
Changes from v2

- Remove useless empty lines :)
- Use git repo instead of tarball for source

ziyao233:
Changes from v3

- move `CHKUPDATE` to correct place

ziyao233:
changes from v4

- make `PKGDES` descriptive

OriginCode (approved):

eatradish (approved):

stdmnpkg (approved):

xtexx (approved):

PR #7628, opened 2024-08-10
systemd: disable systemd-resolved in the default preset

OriginCode:
Topic Description
-----------------

- systemd: disable systemd-resolved in the default preset
    Signed-off-by: Kaiyang Wu <origincode@aosc.io>

Package(s) Affected
-------------------

- systemd: 1:256.4-1

Security Update?
----------------

No

Build Order
-----------

```
#buildit systemd
```

Test Build(s) Done
------------------

**Primary Architectures**

- [x] AMD64 `amd64`
- [x] AArch64 `arm64`

**Secondary Architectures**

- [x] Loongson 3 `loongson3`
- [x] RISC-V 64-bit `riscv64`

### DIFF
--- a/app-admin/systemd/autobuild/patches/0001-sleep.conf-AllowHibernation-no-by-default.patch
+++ b/app-admin/systemd/autobuild/patches/0001-sleep.conf-AllowHibernation-no-by-default.patch
@@ -22,5 +22,5 @@ index 98430348a7..41534e6c15 100644
  #AllowHybridSleep=yes
  #SuspendState=mem standby freeze
 -- 
-2.45.2
+2.46.0
 
--- a/app-admin/systemd/autobuild/patches/0002-presets-disable-systemd-networkd-systemd-networkd-wa.patch
+++ b/app-admin/systemd/autobuild/patches/0002-presets-disable-systemd-networkd-systemd-networkd-wa.patch
@@ -1,19 +1,19 @@
-From 81d7c36800b5bd4c875352a9af84acb7b8d6fcc9 Mon Sep 17 00:00:00 2001
+From 4bc1c27e4ced946439da303a99978bd33449affa Mon Sep 17 00:00:00 2001
 From: Kaiyang Wu <self@origincode.me>
 Date: Fri, 19 Jul 2024 18:38:48 -0700
-Subject: [PATCH 2/3] presets: disable systemd-networkd and
- systemd-networkd-wait-online by default
+Subject: [PATCH 2/3] presets: disable systemd-networkd,
+ systemd-networkd-wait-online and systemd-resolved by default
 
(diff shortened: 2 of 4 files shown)


Icenowy:
Vote against this. It confuses ordinary users.

KexyBiscuit:
# Rationale

- **systemd-resolved** doesn't work well with DNS hijacking features of **Clash Premium** and **sing-box**.
  - **sing-box** tried, but still the same case.
- Even if `/etc/resolv.conf` is managed by other packages, like **NetworkManager**, **systemd-nspawn** will still use the stub `resolv.conf` file managed by `systemd-resolved.service` (specifically: `/run/systemd/resolve/stub-resolv.conf`) inside of the container, if `systemd-resolved.service` is running.
- In some cases, like some Buildbots using **sing-box**, this combination will just cause the DNS query to fail, which is still easy to rule out. But in rare configurations, specifically Buildbots managed by @KexyBiscuit using **Clash Premium**, this combination will cause some of the DNS queries to get fake-ip records, and some to get real records, which would be not ideal and hard to debug.

# Solutions

## Disable **systemd-resolved** in the default preset of AOSC OS. Also mask the service if needed.

### Pro-s

- Works well with DNS hijacking applications, which is useful under special network environments.
- `/etc/resolv.conf` will be managed by **NetworkManager** alone, providing stable and predictable behaviors.

### Con-s

- **systemd-nspawn** could only copy or bind mount the host's `/etc/resolv.conf` file into the container. If it's copied, no changes from the host's `/etc/resolv.conf` file could be detected or applied. If it's bind mounted, when the host's `/etc/resolv.conf` file is replaced, the container would lost its `/etc/resolv.conf` file.
- Disabling systemd-provided services by default isn't a good idea in general.

## Only Disable **systemd-resolved** on Buildbots with special network environments. Also mask the service if needed.

### Pro-s

- Solves the problem.

### Con-s

- Inconsistent network configuration among the Buildbots and other devices running AOSC OS that are managed by AOSC, could be prone to misconfiguration.
- Users outside of AOSC would still be facing the same problem, documentation work might be needed.

## Try to figure out why DNS hijacking applications won't work with **systemd-resolved**.

### Pro-s

- Solves the problem in the long term.
- Potential upstream contributions.

### Con-s

- Takes time, might takes _a lot of_ time.
- Some DNS hijacking applications are either proprietary or end-of-life. Some are both.

KexyBiscuit:
Closing this pull request for now, though patching `90-systemd.preset` sounds a bad idea, there're already **aosc-os-presets-*** packages available, and the modifications should happen there.

@OriginCode please open a new pull request, disable **systemd-networkd** related services in **aosc-os-presets-*** packages, probably the **-base** one, and drop the patch from **systemd**.

PR #7663, opened 2024-08-12
systemd: disable tint-by-default in *spawn and in systemd-run

MingcongBai:
Topic Description
-----------------

- systemd: disable tint-by-default in *spawn and in systemd-run
    Let's face it, it's weird and we have much better ways to notify the user
    about a system running in a container. Change the default behavior (tint by
    default) to only tint when the user specifies a --background= parameter.

Package(s) Affected
-------------------

- systemd: 1:256.4-1

Security Update?
----------------

No

Build Order
-----------

```
#buildit systemd
```

Test Build(s) Done
------------------

**Primary Architectures**

- [x] AMD64 `amd64`
- [x] AArch64 `arm64`

**Secondary Architectures**

- [x] Loongson 3 `loongson3`
- [x] RISC-V 64-bit `riscv64`


MingcongBai:
Background works, but `--background` seems not to function properly (no matter what color was specified, the background was always tinted blue).

图片

app-admin/systemd/autobuild/patches/0002-presets-disable-systemd-networkd-systemd-networkd-wa.patch
```diff
@@ -23,7 +23,11 @@ index da6b9805fd..dcbcd56a7f 100644
 +disable systemd-networkd-wait-online.service
  enable systemd-nsresourced.socket
  enable systemd-pstore.service
- enable systemd-resolved.service
+-enable systemd-resolved.service
++disable systemd-resolved.service
```
KexyBiscuit: Why is this patch still around?

KexyBiscuit (changes requested):

app-admin/systemd/autobuild/patches/0002-presets-disable-systemd-networkd-systemd-networkd-wa.patch
KexyBiscuit: @OriginCode No consensus was made around this, please revert this patch. Also, move **systemd-networkd** presets to **aosc-os-presets-base** package.

KexyBiscuit: @MingcongBai patches have been updated, please rebase this.

KexyBiscuit (approved):

OriginCode:
LGTM on amd64:
圖片

PR #7692, opened 2024-08-14
grub: Add support for forcing EFI installation to the removable media path

shankerwangmiao:
Topic Description
-----------------

- grub: Add support for forcing EFI installation to the removable media path
    Add an extra option to grub-install "--force-extra-removable". On EFI
    platforms, this will cause an extra copy of the grub-efi image to be
    written to the appropriate removable media patch
    /boot/efi/EFI/BOOT/BOOT$ARCH.EFI as well. This will help with broken
    UEFI implementations where the firmware does not work when configured
    with new boot paths.

Package(s) Affected
-------------------

- grub: 2:2.12+unifont15.1.04-6

Security Update?
----------------

No

Build Order
-----------

```
#buildit grub
```

Test Build(s) Done
------------------

**Primary Architectures**

- [x] AMD64 `amd64`
- [x] AArch64 `arm64`

**Secondary Architectures**

- [x] RISC-V 64-bit `riscv64`


MingcongBai (approved):
Tested good on AMD64, LoongArch.

PR #7880, opened 2024-09-03
x11-base: downrank all xf86 drivers

Cyanoxygen:
Topic Description
-----------------

- x11-base: downrank all xf86 drivers to PKGRECOM
    - This allows users to troubleshoot display issues without breaking the
    system.
    - VMware output driver tends to lag if 3D acceleration is enabled.
    Removing the driver from the dependency list, and use the modesetting
    driver instead.
    - VBox output driver seems to be unused, since modesetting driver is
    utilized for both VBoxVGA and VBoxSVGA graphics, thus removing from
    the dependency list.

Package(s) Affected
-------------------

- x11-base: 3

Security Update?
----------------

No

Build Order
-----------

```
#buildit x11-base
```

Test Build(s) Done
------------------

**Primary Architectures**

- [x] AMD64 `amd64`
- [x] AArch64 `arm64`

**Secondary Architectures**

- [x] Loongson 3 `loongson3`
- [x] RISC-V 64-bit `riscv64`


MingcongBai (approved):
Tested good on AMD64:

图片

PR #7995, opened 2024-09-13
cscope: new, 15.9

AydenMeng:
Topic Description
-----------------

cscope: new, 15.9

Package(s) Affected
-------------------

None

Security Update?
----------------

No

Test Build(s) Done
------------------

**Primary Architectures**

- [ ] AMD64 `amd64`
- [ ] AArch64 `arm64`

**Secondary Architectures**

- [ ] Loongson 3 `loongson3`
- [ ] RISC-V 64-bit `riscv64`

### DIFF
--- a/app-devel/cscope/autobuild/defines
+++ b/app-devel/cscope/autobuild/defines
@@ -1,0 +1,4 @@
+PKGNAME=cscope
+PKGSEC=devel
+PKGDEP="ncurses glibc"
+PKGDES="Source code browsing tool based on interactive text screen"
--- a/app-devel/cscope/spec
+++ b/app-devel/cscope/spec
@@ -1,0 +1,4 @@
+VER=15.9
+SRCS="tbl::https://sourceforge.net/projects/cscope/files/cscope/v$VER/cscope-$VER.tar.gz"
+CHKSUMS="sha256::c5505ae075a871a9cd8d9801859b0ff1c09782075df281c72c23e72115d9f159"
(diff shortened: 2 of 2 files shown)


app-devel/cscope/autobuild/defines
```diff
@@ -0,0 +1,6 @@
+PKGNAME=cscope
+PKGSEC=devel
+PKGDEP="ncurses glibc"
+PKGDES="Source code browsing tool based on interactive text screen"
+
+ABSHADOW=0
```
eatradish: Please add a comment explaining why `ABSHADOW=0`

eatradish (dismissed):

app-devel/cscope/autobuild/defines
AydenMeng: The Makefile generated by configure installs the binary to the system directory by default.
Do I need to add this sentence to the define file? By the way, I have grep the define files of other projects and found no similar comments.

eatradish: @AydenMeng 

> By the way, I have grep the define files of other projects and found no similar comments.

A lot of this is due to the history of irregularly introduced packages, and this is something that we don't have time to deal with yet

AydenMeng: For some reasons, I just prepared to add commet in the patch, ~~but I noticed that the patch has been merged~~, I should have prepared it earlier, I am very sorry.

MingcongBai (approved):
Alrighty, ship it!

PR #8005, opened 2024-09-14
openssh: guard against empty invalid or empty keys

MingcongBai:
Topic Description
-----------------

- openssh: guard against empty invalid or empty keys
    Old systems with /etc/ssh/sshd_host_dsa_key{,.pub} would trigger re-generation
    unconditionally. This is a bug - these keys were no longer recognised by
    ssh-keygen, but it would return an empty output (as opposed to any sort of
    error message) in stdin, giving $_key_fingerprint an empty value.
    This empty value would then return a true value as it was grep'd against the
    revoked lists, causing the postinst script to think that the system contained
    a vulnerable key when it really didn't.
    Check for empty values to prevent this.

Package(s) Affected
-------------------

- openssh: 9.8p1-6

Security Update?
----------------

No

Build Order
-----------

```
#buildit openssh
```

Test Build(s) Done
------------------

**Primary Architectures**

- [x] AMD64 `amd64`
- [x] AArch64 `arm64`

**Secondary Architectures**

- [x] Loongson 3 `loongson3`
- [x] RISC-V 64-bit `riscv64`

### DIFF
--- a/app-network/openssh/autobuild/postinst
+++ b/app-network/openssh/autobuild/postinst
@@ -2,6 +2,10 @@ _match_key() {
 	[ -e "/etc/ssh/$1" ] || return 1
 	local _key_sha1
 	_key_sha1="$(sha1sum "/etc/ssh/$1" | cut -d ' ' -f 1)"
+	# Guard against empty or invalid keys - empty greps would return 0.
+	if [ -z "$_key_sha1" ]; then
+		return 1
+	fi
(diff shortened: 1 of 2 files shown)


MingcongBai:
OpenSSH update installed without re-generation routine on an older installation with `/etc/ssh/ssh_host_dsa_key*` (now considered invalid by `ssh-keygen`):

图片

MingcongBai:
... and when injected with a known leaked key:

图片

MingcongBai:
... and after a re-generation:

图片

Cyanoxygen (approved):

PR #8495, opened 2024-11-03
keyd: new, 2.5.0

poscat0x04:
Topic Description
-----------------

Adds keyd

Package(s) Affected
-------------------

- keyd

Security Update?
----------------

No

Test Build(s) Done
------------------

**Primary Architectures**

- [x] AMD64 `amd64`


app-utils/keyd/autobuild/beyond
```diff
@@ -0,0 +1,6 @@
+abinfo "Installing helper scripts"
+
```
MingcongBai: Drop this newline.

app-utils/keyd/autobuild/build
```diff
@@ -0,0 +1,7 @@
+abinfo "Building keyd"
+
```
MingcongBai: Drop this newline.

app-utils/keyd/autobuild/build
```diff
@@ -0,0 +1,7 @@
...
+make PREFIX=/usr
+
+abinfo "Installing keyd"
+
```
MingcongBai: Ditto.

app-utils/keyd/autobuild/build
```diff
@@ -0,0 +1,7 @@
...
+make PREFIX=/usr DESTDIR="$PKGDIR/" install
```
MingcongBai: Why the slash after `$PKGDIR`?

app-utils/keyd/autobuild/overrides/usr/lib/sysusers.d/keyd.conf
```diff
@@ -0,0 +1 @@
+g keyd
```
MingcongBai, line 1: Seems like the user would need to configure their group property to make use of keyd? If so, please introduce a [debconf prompt template](http://www.fifi.org/doc/debconf-doc/tutorial.html) to notify the user.

You might reference `app-devel/intel-oneapi-basekit` for how this is done.

MingcongBai (changes requested):

app-utils/keyd/autobuild/overrides/usr/lib/sysusers.d/keyd.conf
poscat0x04: It's used for an experimental feature. Personally I think it's a niche feature and shouldn't bother a normal user for it.

app-utils/keyd/autobuild/overrides/usr/lib/sysusers.d/keyd.conf
MingcongBai, line 1: You might also need a `postinst` script to create this group:

```
systemd-sysusers keyd.conf
```

MingcongBai (changes requested):

MingcongBai (approved):

PR #8737, opened 2024-11-21
plasma-desktop: drop possible mis-set Flat mouse acceleration profile

MingcongBai:
Topic Description
-----------------

plasma-desktop: drop possible mis-set Flat mouse acceleration profile

Revert upstream commit fb8a655a7376 ("If no config option for accel profile is selected, use default from Xorg").

With libinput 1.26.2, a "flat" acceleration profile was set by default for all input devices. This is probably not correct as libinput still indicates that "adaptive" is the default.

Having a "flat" default makes TrackPoints painfully unusable after login (whereas in SDDM they work as expected).

The user may still select a "flat" profile via the KCM module, save for the issue where their X configuration may not match with that of KDE's on first boot (before this reverted commit). This is really much less of a problem for most users.

Package(s) Affected
-------------------

- plasma-desktop: 5.27.11-3

Security Update?
----------------

No

Build Order
-----------

```
#buildit plasma-desktop
```

Test Build(s) Done
------------------

**Primary Architectures**

- [x] AMD64 `amd64`
- [x] AArch64 `arm64`

**Secondary Architectures**

- [x] Loongson 3 `loongson3`
- [x] RISC-V 64-bit `riscv64`

### DIFF
--- a/desktop-kde/plasma-desktop/autobuild/patches/0001-Revert-If-no-config-option-for-accel-profile-is-sele.patch
+++ b/desktop-kde/plasma-desktop/autobuild/patches/0001-Revert-If-no-config-option-for-accel-profile-is-sele.patch
@@ -1,0 +1,122 @@
+From 86f759380ea4990d4c05cb940cce82d7c433c541 Mon Sep 17 00:00:00 2001
+From: Mingcong Bai <jeffbai@aosc.io>
+Date: Thu, 21 Nov 2024 14:56:43 +0800
+Subject: [PATCH 1/2] Revert "If no config option for accel profile is
+ selected, use default from Xorg"
+
+With libinput 1.26.2, a "Flat" acceleration profile was set by default for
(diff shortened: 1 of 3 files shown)


MingcongBai:
This is not an ideal fix but "the lesser of two evil" before we could figure this out with the upstream. Bug reported at https://bugs.kde.org/show_bug.cgi?id=496531.

Tested good on my Lenovo ThinkPad X280 and my ThinkBook 14 G6+ AHP, as well as a mouse. The slow-cursor issue had been fixed on the ThinkPad and no adverse effect observed on the ThinkBook, nor a mouse. The KCM settings to toggle between "adaptive" and "flat" profiles still works well.

I'd say it's a win until we find a more well-rounded solution.

OriginCode:
Tested settings successfully applied on loongarch64, though I personally didn't feel any big difference.

OriginCode (approved):

PR #8994, opened 2024-12-18
mesa: fix GLX w/o HW accel

Icenowy:
Topic Description
-----------------

- mesa: remove patches that revert dril-related commits
    With these patches applied and no HW acceleration is available, Xorg
    seems to fail to start GLX at all, and GLX programs will fail to run. In
    addition, the color problem these patches use to try to fix seems to be
    already solved by upstream.
    Drop these patch now.
    Signed-off-by: Icenowy Zheng <uwu@icenowy.me>

Package(s) Affected
-------------------

- mesa: 1:24.3.0+dxheaders1.614.1-1

Security Update?
----------------

No

Build Order
-----------

```
#buildit mesa
```

Test Build(s) Done
------------------

**Primary Architectures**

- [x] AMD64 `amd64`
- [x] AArch64 `arm64`

**Secondary Architectures**

- [x] Loongson 3 `loongson3`
- [x] RISC-V 64-bit `riscv64`


Icenowy:
Tested on QEMU with `-std vga` (bochs-drm).

GLX does not work at all before removing these patches, and works (and color is okay) after.

lunzima:
QQ20241218-211911
This patch fixes my Hyper-V setup which had broken glxinfo/glxgears and "Dog chewed" KDE6 taskbar/menus.

MingcongBai (approved):

PR #9090, opened 2024-12-28
systemd: Fix booting problem with QEMU user emulation

Cyanoxygen:
Topic Description
-----------------

- systemd: fix credential setup issue with QEMU user emulation
    Systemd tries to fork() with CLONE_NEWNS, which is not supported by QEMU
    user emulation, to set up credentials. Previously few systemd units use
    ImportCredential=, so the container can still boot with QEMU user
    emulation. Now almost all units use it, rendering the container
    unbootable.
    This fix tries to fork() again without CLONE_NEWNS if the returned value
    is -EINVAL, similar as [1].
    [1]: https://github.com/systemd/systemd/pull/28954

Package(s) Affected
-------------------

- systemd: 1:256.7-1

Security Update?
----------------

No

Build Order
-----------

```
#buildit systemd
```

Test Build(s) Done
------------------

**Primary Architectures**

- [x] AMD64 `amd64`
- [x] AArch64 `arm64`

**Secondary Architectures**

- [x] RISC-V 64-bit `riscv64`


Cyanoxygen:
Testing
--------

arm64 container booted successfully with QEMU user emulation on an amd64 host machine:

image
image

MingcongBai (approved):

PR #9170, opened 2025-01-04
yaml-cpp-0.7: new, 0.7.0

Shyliuli:
Topic Description
-----------------
 - yaml-cpp-0.7: new, 0.7.0
 

Package(s) Affected
-------------------

 - yaml-cpp-0.7: 0.7.0

Security Update?
----------------

No

Build Order
-----------

```
#buildit yaml-cpp-0.7
```

Test Build(s) Done
------------------

**Primary Architectures**

- [x] AMD64 `amd64`
- [x] AArch64 `arm64`

**Secondary Architectures**

- [x] Loongson 3 `loongson3`
- [x] RISC-V 64-bit `riscv64`

### DIFF
--- a/runtime-common/yaml-cpp-0.7/autobuild/beyond
+++ b/runtime-common/yaml-cpp-0.7/autobuild/beyond
@@ -1,0 +1,3 @@
+abinfo "Removing unused gmock and gtest files ..."
+rm -rv "$PKGDIR"/usr/lib/legacy/yaml-cpp-0.7/cmake/GTest
+rm -rv "$PKGDIR"/usr/lib/legacy/yaml-cpp-0.7/pkgconfig/{gmock,gtest}*.pc
--- a/runtime-common/yaml-cpp-0.7/autobuild/defines
+++ b/runtime-common/yaml-cpp-0.7/autobuild/defines
@@ -1,0 +1,13 @@
+PKGNAME=yaml-cpp-0.7
+PKGSEC=libs
+PKGDEP="boost"
+BUILDDEP="cmake"
+PKGDES="YAML parser and emitter in C++ (legacy 0.7 runtime)"
+
+ABTYPE=cmake
+CMAKE_AFTER=(
+             -DBUILD_SHARED_LIBS=ON
+             -DYAML_BUILD_SHARED_LIBS=ON
+             -DCMAKE_INSTALL_LIBDIR=lib/legacy/yaml-cpp-0.7
+             -DCMAKE_INSTALL_INCLUDEDIR=include/yaml-cpp-0.7
+)
--- a/runtime-common/yaml-cpp-0.7/spec
+++ b/runtime-common/yaml-cpp-0.7/spec
@@ -1,0 +1,4 @@
+VER=0.7.0
+SRCS="git::commit=tags/yaml-cpp-$VER::https://github.com/jbeder/yaml-cpp/"
+CHKSUMS="SKIP"
+CHKUPDATE="anitya::id=376236"
--- a/runtime-common/yaml-cpp-0.7/autobuild/overrides/etc/ld.so.conf.d/yaml-cpp-0.7.conf
+++ b/runtime-common/yaml-cpp-0.7/autobuild/overrides/etc/ld.so.conf.d/yaml-cpp-0.7.conf
@@ -1,0 +1,1 @@
+/usr/lib/legacy/yaml-cpp-0.7
--- a/runtime-common/yaml-cpp-0.7/autobuild/postinst
+++ b/runtime-common/yaml-cpp-0.7/autobuild/postinst
@@ -1,0 +1,2 @@
+echo "Refreshing dynamic library cache ..."
+ldconfig


runtime-common/yaml-cpp-0.7/autobuild/overrides/etc/ls.so.conf.d/yaml-cpp-0.7.conf
MingcongBai, line 1: Missing newline at end-of-file (Visual Studio Code?).

runtime-common/yaml-cpp-0.7/autobuild/beyond
```diff
@@ -0,0 +1,2 @@
+rm -rf "$PKGDIR"/usr/include/{gmock,gtest}
```
MingcongBai: Per the [Styling Manual](https://wiki.aosc.io/zh/developer/packaging/package-styling-manual/):

- Add an abinfo indicating what is currently underway (`abinfo "Removing unused gmock and gtest files ..."`) - though you should really double-check why they were installed in the first place.
- Missing verbosity, you should be using `-v`.
- No `-f` please, as it makes it an unpredictable action.

runtime-common/yaml-cpp-0.7/autobuild/defines
```diff
@@ -0,0 +1,11 @@
...
+PKGSEC=devel
+PKGDEP="boost"
+BUILDDEP="cmake"
+PKGDES="YAML parser and emitter in C++, written around the YAML 1.2 spec"
+ABTYPE=cmake
+CMAKE_AFTER="-DBUILD_SHARED_LIBS=ON \
+             -DYAML_BUILD_SHARED_LIBS=ON \
+             -DCMAKE_INSTALL_LIBDIR=lib/legacy/yaml-cpp-0.7 \
+             -DCMAKE_INSTALL_INCLUDEDIR=include/yaml-cpp-0.7"
+PKGBREAK="librime<=1:1.2.9-1 mongodb<=3.6.3"
```
MingcongBai: What? This package was never introduced so why would it break anything?

runtime-common/yaml-cpp-0.7/autobuild/defines
```diff
@@ -0,0 +1,11 @@
+PKGNAME=yaml-cpp-0.7
+PKGSEC=devel
+PKGDEP="boost"
+BUILDDEP="cmake"
+PKGDES="YAML parser and emitter in C++, written around the YAML 1.2 spec"
```
MingcongBai: YAML parser and emitter in C++ (legacy 0.7 runtime)

runtime-common/yaml-cpp-0.7/autobuild/defines
```diff
@@ -0,0 +1,11 @@
...
+ABTYPE=cmake
```
MingcongBai, line 7: A new line here will make the file much more readable.

runtime-common/yaml-cpp-0.7/autobuild/defines
```diff
@@ -0,0 +1,11 @@
+PKGNAME=yaml-cpp-0.7
+PKGSEC=devel
```
MingcongBai: Should be `libs`.

runtime-common/yaml-cpp-0.7/spec
```diff
@@ -0,0 +1,4 @@
+VER=0.7.0
+SRCS="https://github.com/jbeder/yaml-cpp/archive/yaml-cpp-$VER.tar.gz"
+CHKSUMS="sha256::43e6a9fcb146ad871515f0d0873947e5d497a1c9c60c58cb102a97b47208b7c3"
+CHKUPDATE="anitya::id=5284"
```
MingcongBai: You would need to create a new Anitya entry for 0.7.x.

MingcongBai (changes requested):

runtime-common/yaml-cpp-0.7/spec
Shyliuli: but how to do..?

runtime-common/yaml-cpp-0.7/autobuild/beyond
Shyliuli: OK!
and by the way,this is copied from yaml-cpp(0.6.3)...I reckon that We may should use 
lib/legacy/yaml-cpp-0.7
and
include/yaml-cpp-0.7

MingcongBai: > OK! and by the way,this is copied from yaml-cpp(0.6.3)...I reckon that We may should use lib/legacy/yaml-cpp-0.7 and include/yaml-cpp-0.7

I think you would be correct.

runtime-common/yaml-cpp-0.7/autobuild/defines
```diff
@@ -0,0 +1,10 @@
+PKGNAME=yaml-cpp-0.7
+ABTYPE=cmake
```
MingcongBai: Weird placement, can you move this to above CMAKE_AFTER. Also, add a blank line after PKGDES.

MingcongBai (changes requested):

runtime-common/yaml-cpp-0.7/spec
```diff
@@ -0,0 +1,4 @@
+VER=0.7.0
+SRCS="https://github.com/jbeder/yaml-cpp/archive/yaml-cpp-$VER.tar.gz"
```
stdmnpkg: use git instead of tarball

runtime-common/yaml-cpp-0.7/autobuild/defines
```diff
@@ -0,0 +1,11 @@
+PKGNAME=yaml-cpp-0.7
+PKGSEC=libs
+PKGDEP="boost"
+BUILDDEP="cmake"
+PKGDES="YAML parser and emitter in C++ (legacy 0.7 runtime)"
+
+ABTYPE=cmake
+CMAKE_AFTER="-DBUILD_SHARED_LIBS=ON \
```
stdmnpkg: this is just a suggestion, you can write this as array

```bash
CMAKE_AFTER=(
    -DBUILD_SHARED_LIBS=ON
    -DYAML_BUILD_SHARED_LIBS=ON
    ......
)
```

stdmnpkg (approved):

stdmnpkg (changes requested):
sorry, drunk too much

OriginCode:
~~Why adding a legacy version when there's no package depending on this version?~~

Nevermind, for 3rd-party packages.

runtime-common/yaml-cpp-0.7/autobuild/overrides/etc/ls.so.conf.d/yaml-cpp-0.7.conf
OriginCode, line 1: There's a typo in the path.

OriginCode (dismissed):

Shyliuli:
> ~Why adding a legacy version when there's no package depending on this version?~
> 
> Nevermind, for 3rd-party packages.

I am working on the One Student One Chip project, which involves compiling yoyos, requiring libraries of this version.
I noticed that the package on AOSC is still at version 0.6.3, and I thought about helping to update it, but the upstream has already reached version 0.8.
After discussing with Mr. Bai Mingcong, we decided to retain version 0.7 and update the package to version 0.8.

MingcongBai (approved):

stdmnpkg (approved):
lgtm, since i have no program require this, paste some test result if possible

PR #9392, opened 2025-01-21
linux-kernel-lemote: drop, orphaned

MingcongBai:
Topic Description
-----------------

- linux\-kernel\-lemote: drop, orphaned
    Now that the mainline kernel is what we target to support \(and it seems to
    work well enough\), retire the Lemote 5.4 kernel which aged like yogurt.

Package(s) Affected
-------------------

Security Update?
----------------

No

Build Order
-----------

```
#buildit linux-kernel-lemoet linux+kernel+lemote
```

Test Build(s) Done
------------------

**Primary Architectures**

- [ ] AMD64 `amd64`
- [ ] AArch64 `arm64`

**Secondary Architectures**

- [ ] Loongson 3 `loongson3`
- [ ] RISC-V 64-bit `riscv64`


KexyBiscuit (approved):
Never conducted real due diligence for this tree... Great job @MingcongBai.

PR #9531, opened 2025-02-02
binutils: update to 2.44

MingcongBai:
Topic Description
-----------------

- gcc+cross: depend on binutils
    Binutils now comes with support for all targets.
- binutils: update to 2.44
    - Enable a unified build for all targets.
    - Revise sed invocations into patches.
    - Track patches at AOSC-Tracking/binutils-gdb @ aosc/binutils-2_44
    \(HEAD: 072457058d06965a634c35c44b6ecc3c62609e56\).

Package(s) Affected
-------------------

- binutils: 2.44
- gcc+cross-amd64: 13.2.0+glibc2.38+6-3
- gcc+cross-arm64: 13.2.0+glibc2.38+6-3
- gcc+cross-loongarch64: 13.2.0+glibc2.38+6-3
- gcc+cross-loongson3: 13.2.0+glibc2.38+6-3
- gcc+cross-ppc64el: 13.2.0+glibc2.38+6-3
- gcc+cross-riscv64: 13.2.0+glibc2.38+6-3

Security Update?
----------------

No

Build Order
-----------

```
#buildit binutils gcc+cross
```

Test Build(s) Done
------------------

**Primary Architectures**

- [ ] AMD64 `amd64`
- [ ] AArch64 `arm64`

**Secondary Architectures**

- [ ] Loongson 3 `loongson3`
- [ ] RISC-V 64-bit `riscv64`


xry111:
I've some doubts on the patches.  `-O2` shouldn't have an effect on the preprocessing (unless some header is checking `__OPTIMIZE__` but then that header would be completely stupid) and we shouldn't need to raise `ELF_COMMONPAGESIZE` as long as `ELF_MAXPAGESIZE` is correct.

LoongArch also has `ELF_COMMONPAGESIZE` = 0x4000 and `ELF_MAXPAGESIZE` = 0x10000, and per various reports the page size issues are fixed after @xen0n raised `ELF_MAXPAGESIZE` (w/o changing `ELF_COMMONPAGESIZE`).

MingcongBai:
> I've some doubts on the patches. `-O2` shouldn't have an effect on the preprocessing (unless some header is checking `__OPTIMIZE__` but then that header would be completely stupid) and we shouldn't need to raise `ELF_COMMONPAGESIZE` as long as `ELF_MAXPAGESIZE` is correct.
> 
> LoongArch also has `ELF_COMMONPAGESIZE` = 0x4000 and `ELF_MAXPAGESIZE` = 0x10000, and per various reports the page size issues are fixed after @xen0n raised `ELF_MAXPAGESIZE` (w/o changing `ELF_COMMONPAGESIZE`).

You are probably right. These were originally `sed` routines in the prepare script, which as you know, are now discouraged - they came from Fedora, so it might be a good idea to dig into how they came to that hack (just as a note, a few years back, we had a pretty bad practice of syncing patches and hacks from other distributions without due diligence).

MingcongBai:
Superseded.

PR #9533, opened 2025-02-03
git: build with Meson instead of GNU Make

KexyBiscuit:
Topic Description
-----------------

- git: lint build scripts...
    Signed-off-by: Kexy Biscuit <kexybiscuit@aosc.io>
- git: split git-gui; use a dedicated BLDDIR; fix autobuild crash
    - By making a copy of the autobuild directory to the new SRCDIR.
- git: use a dedicated PKGDIR and BLDDIR
    - This does not require a cleanup of BLDDIR.
- git: gitk: use a dedicated PKGDIR and BLDDIR; fix autobuild crash
    - By making a copy of autobuild directory to the new SRCDIR.
- gitk: disable dbg package generation
    Signed-off-by: Kexy Biscuit <kexybiscuit@aosc.io>
- git: fix source directory for meson template
    Signed-off-by: Kexy Biscuit <kexybiscuit@aosc.io>
- git: clean up build directory for gitk
    Signed-off-by: Kexy Biscuit <kexybiscuit@aosc.io>
- git: fix install command parameters
    Signed-off-by: Kexy Biscuit <kexybiscuit@aosc.io>
- git: split into three packages...
    - Package \`git\` is actually git-core.
    - Bash and zsh completions are now installed.
    - gitk and git-gui are now built as separate packages.
    Signed-off-by: Kexy Biscuit <kexybiscuit@aosc.io>
- git: add patch to fix meson building
    - Track patches at https://github.com/AOSC-Tracking/git.git aosc/v2.49.0-rc1, current HEAD is 085c64ea410e3331d4a39caacdc546ac64ea3271.
    Signed-off-by: Kexy Biscuit <kexybiscuit@aosc.io>

... and 4 more commits

Package(s) Affected
-------------------

- git: 2.49.0-0.1meson2
- git-gui: 2.49.0-0.1meson2
- gitk: 2.49.0-0.1meson2

Security Update?
----------------

No

Build Order
-----------

```
#buildit git
```

Test Build(s) Done
------------------

**Primary Architectures**

- [x] AMD64 `amd64`
- [x] AArch64 `arm64`

**Secondary Architectures**

- [x] Loongson 3 `loongson3`
- [x] RISC-V 64-bit `riscv64`


KexyBiscuit:
TODOs:
- ~~`git-gui` and `gitk` are missing from Meson.~~
- ~~Shell completions are not copied.~~
- ~~`/usr/libexec` is used instead of `/usr/lib`, is that correct?~~ Yes.
- ~~`git-subtree` is not built since it's in `contrib`, everything in `contrib` other than completion has no Meson build file.~~
- Maybe I have missed something... The list is just too long.

Will revisit later... Maybe Git 3.0?

xry111:
> * `/usr/libexec` is used instead of `/usr/lib`, is that correct?

FYI we (BLFS) are using `/usr/libexec` (with autotools).  Several years ago we had a debate about `/usr/libexec/$pkg` vs `/usr/lib/$pkg` and the conclusion was just using the upstream default.

MingcongBai:
> > * `/usr/libexec` is used instead of `/usr/lib`, is that correct?
> 
> FYI we (BLFS) are using `/usr/libexec` (with autotools). Several years ago we had a debate about `/usr/libexec/$pkg` vs `/usr/lib/$pkg` and the conclusion was just using the upstream default.

I do second that /usr/libexec will work just fine.

eli-schwartz:
> - `/usr/libexec` is used instead of `/usr/lib`, is that correct?

This isn't actually a fair comparison :) libexec is the default upstream but the GNU Make build set this:

```
    gitexecdir='/usr/lib/git-core'
```

It is technically possible to override this with meson too, simply configure via `--libexecdir=/usr/lib`. Whether it is worth the deviation is up to you.



> `git-subtree` is not built since it's in `contrib`, everything in `contrib` other than completion has no Meson build file.

Everything in `contrib` was never built by the Makefile either. I think the ultimate plan is to expand the meson files to handle contrib for the first time, though. :)

KexyBiscuit:
@eli-schwartz Thank you for pointing out! The "TODO" list was meant as a To-Do and RFC for the community, not an actual bug report for the upstream :-) Anyway we'll change to use `libexec` after changing to Meson build.

`contrib` received proper Meson build system support in <https://lore.kernel.org/all/20250117-b4-pks-build-subtree-v1-0-03c2ed6cc42e@pks.im/>. Will take a look at this later this week... Was OOO recently so sorry for the late reply.

PR #9670, opened 2025-02-13
openssl-1.1: update to 1.1.1zb+p3

stdmnpkg:
Topic Description
-----------------

- openssl-1.1: update to 1.1.1zb+p3
    Switched to https://github.com/kzalewski/openssl-1.1.1 for security update backported from OpenSSL 3.
    OpenSSL no longer provide free security update to OpenSSL 1.1.1 after 1.1.1w.

Package(s) Affected
-------------------

- openssl-1.1: 1.1.1zb+p3

Security Update?
----------------

No

Build Order
-----------

```
#buildit openssl-1.1
```

Test Build(s) Done
------------------

**Primary Architectures**

- [x] AMD64 `amd64`
- [x] AArch64 `arm64`

**Secondary Architectures**

- [x] Loongson 3 `loongson3`
- [x] RISC-V 64-bit `riscv64`

### DIFF
--- a/runtime-cryptography/openssl-1.1/autobuild/build
+++ b/runtime-cryptography/openssl-1.1/autobuild/build
@@ -9,6 +9,8 @@ elif ab_match_arch "i486" ; then
     ARCH_OPTS="linux-x86"
 elif ab_match_arch "loongarch64"; then
     ARCH_OPTS="linux-generic64"
+elif ab_match_arch "loongarch64_nosimd"; then
+    ARCH_OPTS="linux-generic64"
 elif ab_match_arch "+(loongson2f|loongson3)" ; then
     ARCH_OPTS="linux64-mips64"
 elif ab_match_arch "m68k" ; then
--- a/runtime-cryptography/openssl-1.1/autobuild/patches/0001-patch-Configure-for-loongson3.patch.loongson3
+++ b/runtime-cryptography/openssl-1.1/autobuild/patches/0001-patch-Configure-for-loongson3.patch.loongson3
@@ -1,0 +1,25 @@
+From 7bc94b7861c09c5307af4dfcfdb6a3439b4d37bc Mon Sep 17 00:00:00 2001
+From: Student Main <studentmain@aosc.io>
+Date: Wed, 12 Feb 2025 13:27:40 +0800
+Subject: [PATCH] patch Configure for loongson3
+
+---
+ Configure | 2 +-
+ 1 file changed, 1 insertion(+), 1 deletion(-)
+
+diff --git a/Configure b/Configure
+index 78cc15d..d6484d5 100755
+--- a/Configure
++++ b/Configure
+@@ -1248,7 +1248,7 @@ if ($target =~ /linux.*-mips/ && !$disabled{asm}
+         # minimally required architecture flags for assembly modules
+         my $value;
+         $value = '-mips2' if ($target =~ /mips32/);
+-        $value = '-mips3' if ($target =~ /mips64/);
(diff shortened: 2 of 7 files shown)


stdmnpkg:
rationale of switch to this new upstream:

1. some third party security fix is better than no security fix after all
2. [openEuler's fork](https://gitee.com/openeuler/openssl) was also considered, but it introduced some new feature, not a pure security fix
3. this upstream is also [used in slackware](https://slackware.uk/slackware/slackware64-15.0/patches/source/openssl/)
4. author also maintained other open source project for at least 10 years

runtime-cryptography/openssl-1.1/autobuild/build
```diff
@@ -40,7 +40,7 @@ fi
 abinfo "Running Configure ..."
 "$SRCDIR"/Configure --prefix=/usr --openssldir=/etc/ssl --libdir=lib \
             shared zlib ${ARCH_OPTS}\
-            "-Wa,--noexecstack ${CPPFLAGS} ${CFLAGS}"
+            "-Wa,--noexecstack -Wno-implicit-function-declaration ${CPPFLAGS} ${CFLAGS}"
```
MingcongBai: This is still quite suspect, I would recommend digging deeper.

runtime-cryptography/openssl-1.1/autobuild/patches/0001-patch-Configure-for-loongson3.patch.loongson3
```diff
@@ -0,0 +1,25 @@
...
+diff --git a/Configure b/Configure
+index 78cc15d..d6484d5 100755
+--- a/Configure
++++ b/Configure
+@@ -1248,7 +1248,7 @@ if ($target =~ /linux.*-mips/ && !$disabled{asm}
+         # minimally required architecture flags for assembly modules
+         my $value;
+         $value = '-mips2' if ($target =~ /mips32/);
+-        $value = '-mips3' if ($target =~ /mips64/);
++        $value = '-mips64r2' if ($target =~ /mips64/);
```
MingcongBai, line 19: Please make this architecturally conditional - our other mips64el port, loongson2f, is mips3.

MingcongBai: Also, would this be something to do also for openssl (the main 3.x package)? cc @chenx97 @FlyGoat @rickliu2000

MingcongBai (changes requested):

runtime-cryptography/openssl-1.1/autobuild/build
stdmnpkg: should fixed by https://github.com/kzalewski/openssl-1.1.1/pull/7

MingcongBai:
Please package p3 instead with build fix.

MingcongBai:
I think, let's spend more time analyzing the existing fixes. @stdmnpkg has agreed to do a three-way comparison between OpenSSL 3.x, this upstream, and openEuler.

stdmnpkg:
i compared openssl 3, openeuler fork and this one, here's preliminary result

except test and document, there are few patch missing from this upstream.
i will open a pull request to upstream.

some definition's value are different with openeuler's, but i doubt its significance,
application which need these values should already shipped with bundled openssl.

otherwise they are effectively same. and openssl 1.1.1w from this one is same as official one.

[openssl-diff-report.txt](https://github.com/user-attachments/files/19234833/openssl-diff-report.txt)

MingcongBai (dismissed):

MingcongBai (approved):

PR #9757, opened 2025-02-19
procps: generate correct version information

ziyao233:
Topic Description
-----------------

- procps: generate correct version information
    procps tries to generate version information from .tarball-release
    file and git history, or will use "UNKNOWN" if both methods fail.
    Let's create a .tarball-release file from PKGVER to prevent ulities
    from taking "UNKNOWN" as their version string, which is confusing.
    Signed-off-by: Yao Zi <ziyao@disroot.org>

Package(s) Affected
-------------------

- procps: 4.0.5-1

Security Update?
----------------

No

Build Order
-----------

```
#buildit procps
```

Test Build(s) Done
------------------

**Primary Architectures**

- [x] AMD64 `amd64`
- [x] AArch64 `arm64`

**Secondary Architectures**

- [x] Loongson 3 `loongson3`
- [x] RISC-V 64-bit `riscv64`

### DIFF
--- a/app-utils/procps/autobuild/prepare
+++ b/app-utils/procps/autobuild/prepare
@@ -1,0 +1,2 @@
+abinfo "Generating .tarball-version"
+echo "$PKGVER" > "$SRCDIR/.tarball-version"
--- a/app-utils/procps/spec
+++ b/app-utils/procps/spec
@@ -1,4 +1,5 @@
 VER=4.0.5
+REL=1
 SRCS="https://gitlab.com/procps-ng/procps/-/archive/v$VER/procps-v$VER.tar.gz"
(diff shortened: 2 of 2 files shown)


app-utils/procps/autobuild/prepare
```diff
@@ -0,0 +1 @@
+echo "$PKGVER" > "$SRCDIR/.tarball-version"
```
MingcongBai, line 2: Please add an `abinfo` indicating what was being done here.

MingcongBai (changes requested):

MingcongBai:
LGTM:

```
mingcongbai@JellyNote [ ~ ] $ top --version
top，来自 procps-ng 4.0.5
```

app-utils/procps/autobuild/prepare
ziyao233: Done

MingcongBai (approved):

PR #9838, opened 2025-02-24
xorg-server: workaround bad firmware boot VGA reporting on MIPS Loongson devices

MingcongBai:
Topic Description
-----------------

- xorg-server: workaround bad firmware boot VGA reporting on MIPS Loongson devices
    Loongson \(MIPS\) platforms based on \(at least\) the 7A bridge chip has
    faulty firmware that does not export the correct boot VGA device to the
    operating system, causing the X server to mistake the non-connected output
    as the primary match for X output - which then causes the X server to fail
    to probe for the right driver, in turn causing X server to fail at
    startup.
    From our user report, this motherboard/firmware combination is known to
    have this issue \(there is probably more\):
    BM6J74 \(Kunlun 4.00.0024, OEM version LJ74B00BPS\)
    The original patch from Loongson/UOS uses an extra call to
    \`xf86platformPrimary\(\)' to override the result from the above call to
    \`xf86platformProbe\(\)', prompting X to use the first platform device found
    on the device \(the first video card/GPU probed by the kernel \(for
    modesetting or simply as firmware output\).
    This will most certainly break other platforms without this issue, so
    limit this workaround to little-endian MIPS64 with N64 ABI for now \(for
    which AOSC OS only supports Loongson 3 devices\).
    \[Where this patch was first shown to the public...\]
    Link: https://github.com/deepin-community/xorg-server/pull/11

Package(s) Affected
-------------------

- xorg-server: 21.1.15-2

Security Update?
----------------

No

Build Order
-----------

```
#buildit xorg-server
```

Test Build(s) Done
------------------

**Primary Architectures**

- [x] AMD64 `amd64`
- [x] AArch64 `arm64`

**Secondary Architectures**

- [x] Loongson 3 `loongson3`
- [x] RISC-V 64-bit `riscv64`

### DIFF
--- a/runtime-display/xorg-server/autobuild/patches/0001-Use-intel-ddx-only-on-pre-gen4-hw-newer-ones-will-fa.patch
+++ b/runtime-display/xorg-server/autobuild/patches/0001-Use-intel-ddx-only-on-pre-gen4-hw-newer-ones-will-fa.patch
@@ -1,7 +1,7 @@
-From db964f598a52652c6d3d674a81e577ef3f01b46d Mon Sep 17 00:00:00 2001
+From c77be321d01ba91443845d3f5896969e0c2be546 Mon Sep 17 00:00:00 2001
 From: Kexy Biscuit <kexybiscuit@aosc.io>
 Date: Fri, 30 Aug 2024 14:06:28 +0800
-Subject: [PATCH 1/6] Use intel ddx only on pre-gen4 hw, newer ones will fall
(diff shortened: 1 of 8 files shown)


AirFortressIlikara:
[Xorg.0.log](https://github.com/user-attachments/files/18940026/Xorg.0.log)
截图_2025-02-24_18-21-34
我的配置为3A4000+RX580,可以正常进入桌面。

ruaharico (approved):

PR #10353, opened 2025-04-07
u-boot-menu: revert changes for /boot/u-boot

Icenowy:
Topic Description
-----------------

- Revert "u-boot-menu: Enforce /boot/u-boot as the boot filesystem"
    This reverts commit 442ff7f180602a06cff83e68bf6bde669263b24f.
    The u-boot-menu script generates menu for extlinux, which by definition
    should be able to work on an EXT partition. Even if the / is on a
    filesystem that extlinux does not support, /boot could be a dedicated
    EXT4 partition, which is a native supported configuration of u-boot-menu
    upstream.
    If the boot chain enforces a VFAT partition, putting the extlinux
    configuration file \(and thus the kernel\) on it isn't a good idea,
    because of the risk of kernel getting out-of-sync. The partition should
    be mounted to /efi \(or similar path\) instead.
    Signed-off-by: Icenowy Zheng <uwu@icenowy.me>

Package(s) Affected
-------------------

- u-boot-menu: 4.2.3-3

Security Update?
----------------

No

Build Order
-----------

```
#buildit u-boot-menu
```

Test Build(s) Done
------------------

**Primary Architectures**


Icenowy:
Putting a vf2_uEnv.txt to vfat partition 1 and /boot as ext4 on partition 2 seems to work on Orange Pi RV with SD card.

Seems that I need to investigate with NVMe and on VF2 then... (but it seems that the VF2 firmware receives many updates too)

Icenowy:
Tested on VF2 with `U-Boot 2021.10 (Sep 28 2024 - 01:14:56 +0800), Build: jenkins-github_visionfive2_6.6-5` too, VF2 expects vf2_uEnv.txt in vfat partition 3 instead.

Cyanoxygen (approved):

PR #10400, opened 2025-04-09
openssh: update to 10.0p1

MingcongBai:
Topic Description
-----------------

- openssh: update to 10.0p1
    Co-authored-by: 白铭骢 \(Mingcong Bai\) \(@MingcongBai\) <jeffbai@aosc.io>

Package(s) Affected
-------------------

- openssh: 10.0p1

Security Update?
----------------

No

Build Order
-----------

```
#buildit openssh
```

Test Build(s) Done
------------------

**Primary Architectures**

- [x] AMD64 `amd64`
- [x] AArch64 `arm64`

**Secondary Architectures**

- [x] Loongson 3 `loongson3`
- [x] RISC-V 64-bit `riscv64`

### DIFF
--- a/app-network/openssh/spec
+++ b/app-network/openssh/spec
@@ -1,4 +1,4 @@
-VER=9.9p2
+VER=10.0p1
 SRCS="tbl::https://cdn.openbsd.org/pub/OpenBSD/OpenSSH/portable/openssh-$VER.tar.gz"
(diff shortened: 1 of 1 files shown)


MingcongBai:
Tested good on both client and server side... But the reported version is wrong.

image

... Which is a known issue on the upstream side and they won't fix it, so let's just keep it that way (since all documentation will now refer to this version as 10.0p2:

image

OriginCode (approved):

PR #10648, opened 2025-04-26
clash-verge-service: fix get root access security issue

eatradish:
Topic Description
-----------------

- clash-verge-service: fix security issue ...
    ... https://github.com/clash-verge-rev/clash-verge-rev/issues/3428

Package(s) Affected
-------------------

- clash-verge-service: 0+git20250324-2

Security Update?
----------------

No

Build Order
-----------

```
#buildit clash-verge-service
```

Test Build(s) Done
------------------

**Primary Architectures**

- [x] AMD64 `amd64`
- [x] AArch64 `arm64`

**Secondary Architectures**


xmcp:
The patched code is still vulnerable to arbitrary file write (since the attacker can pass the `log_file` parameter) and information disclosure (through endpoints such as `GET /version`).

I can't find a easy way to fix them though, unless it does some authorization to the API.

MingcongBai:
> The patched code is still vulnerable to arbitrary file write (since the attacker can pass the `log_file` parameter) and information disclosure (through endpoints such as `GET /version`).
> 
> I can't find a easy way to fix them though, unless it does some authorization to the API.

Hi, thanks for your review, here's my take on it (please correct me when I'm wrong):

1. The original issue was a local privilege escalation vulnerability.
2. Theoretically, to exploit this after this patch, we would need to replace `/usr/bin/mihomo`, unless the binary was replaced there (which requires root) - the prerequisite to exploitation is therefore contradictory.

Now, to achieve arbitrary file write or read - we would still need to rely on `/usr/bin/mihomo` (that we ship) having either a function or vulnerability that allows it to happen - which, as we recounted, would be to simply set `log_file` as system/service-critical files, such as `/var/spool/cron/root` and `/var/www/html/webshell.php`.

~~Maybe, @eatradish we would also fix (as in presetting) the `log_file` parameter?~~

Ugh, `log_file` is read from configuration, which is user-editable. Taking minimal privilege approach with `setcap` seems reasonable, though:

https://github.com/clash-verge-rev/clash-verge-rev/issues/3428#issuecomment-2832125194

eatradish:
@xmcp Hi, Maybe https://github.com/AOSC-Tracking/clash-verge-service/commit/cdcf25bee9eaa8406f74d01733c25bd16c3fef8d could be enough to solve the problem? (for Linux Distro)

eatradish:
@aosc-buildit-bot build

xmcp:
> The original issue was a local privilege escalation vulnerability

I believe the consequence of this vuln is much more severe than a privilege escalation. Anyone who can POST to 127.0.0.1 can exploit this, for example, browser extensions:

image

Also if the proxy on the server allows to request 127.0.0.1 through it (depending on how proxy rules are configured), remote attackers can achieve RCE through the proxy.

> (The attacker) would be to simply set log_file as system/service-critical files

Yes indeed. It won't be hard to insert some attack payload if the file format is error proof. E.g., if the attacker can insert `\n` into the log, he can write into `/etc/profile` or `/root/.ssh/authorized_keys`. PHP files are also good examples.

> Maybe https://github.com/AOSC-Tracking/clash-verge-service/commit/cdcf25bee9eaa8406f74d01733c25bd16c3fef8d could be enough to solve the problem?

Yes it seems that attackers can no longer exploit the log parameter. I am still concerned about browser extensions / proxy users have access to the clash service (causing information disclosure and DoS), but at least code execution and arbitraty write are impossible now. A long term solution may require the clash-verge people to add proper authorization.

eatradish:
@aosc-buildit-bot build

MingcongBai:
> > The original issue was a local privilege escalation vulnerability
> 
> I believe the consequence of this vuln is much more severe than a privilege escalation. Anyone who can POST to 127.0.0.1 can exploit this, for example, browser extensions:
> 
> image
> 
> Also if the proxy on the server allows to request 127.0.0.1 through it (depending on how proxy rules are configured), remote attackers can achieve RCE through the proxy.
> 
> > (The attacker) would be to simply set log_file as system/service-critical files
> 
> Yes indeed. It won't be hard to insert some attack payload if the file format is error proof. E.g., if the attacker can insert `\n` into the log, he can write into `/etc/profile` or `/root/.ssh/authorized_keys`. PHP files are also good examples.
> 
> > Maybe [AOSC-Tracking/clash-verge-service@cdcf25b](https://github.com/AOSC-Tracking/clash-verge-service/commit/cdcf25bee9eaa8406f74d01733c25bd16c3fef8d) could be enough to solve the problem?
> 
> Yes it seems that attackers can no longer exploit the log parameter. I am still concerned about browser extensions / proxy users have access to the clash service (causing information disclosure and DoS), but at least code execution and arbitraty write are impossible now. A long term solution may require the clash-verge people to add proper authorization.

Got it. I will write it up as an RCE in TUM (Topic Update Manifest) so users will know. Thanks for the insight.

We won't be publishing a security advisory on our home page/Wechat, for obvious reasons.

MingcongBai (dismissed):
LGTM, and TUM: 

图片

MingcongBai (approved):
Revised TUM:

图片

PR #10831, opened 2025-05-08
apr-util: drop Berkeley DB support

MingcongBai:
Topic Description
-----------------

- apr-util: drop Berkeley DB support
    With GCC >= 14, Berkeley DB support no longer builds. We have been using
    GDBM as the default database manager so it should not affect our
    dependency tree.
    Note that Berkeley DB is installed as part of BuildKit so it is always
    available in the build environment. This revision disables this feature
    explicitly.
- apr: refactor apr/autobuild/define

Package(s) Affected
-------------------

- apr: 1.7.5-1
- apr-util: 1.6.3-2

Security Update?
----------------

No

Build Order
-----------

```
#buildit apr apr-util
```

Test Build(s) Done
------------------

**Primary Architectures**

- [x] AMD64 `amd64`
- [x] AArch64 `arm64`

**Secondary Architectures**

- [x] Loongson 3 `loongson3`
- [x] RISC-V 64-bit `riscv64`

### DIFF
--- a/runtime-common/apr-util/autobuild/defines
+++ b/runtime-common/apr-util/autobuild/defines
@@ -1,6 +1,6 @@
 PKGNAME=apr-util
 PKGDES="The Apache Portable Runtime utilities"
-PKGSEC=utils
+PKGSEC=libs
 PKGDEP="apr expat nss openldap openssl"
 PKGSUG="db gdbm mariadb postgresql sqlite unixodbc"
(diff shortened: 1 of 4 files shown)


MingcongBai:
File list looks good as we dropped Berkeley DB support. There should not be other functional differences.

AirFortressIlikara (approved):

PR #11270, opened 2025-06-04
Experimental Loongson3 CPUFreq ABIv1 driver

ziyao233:
Topic Description
-----------------

- linux-kernel: Support cpufreq on Loongson 3 with older SMC firmware
    Introduces the experimental downstream patch for CPUFreq support on
    Loongson 3 machines with older SMC firmware \(CMD_GET_VERSION returns 0
    instead of a value greater than 0\).
    Signed-off-by: Yao Zi <ziyao@disroot.org>

Package(s) Affected
-------------------

- linux-kernel-${__VER}: 6.14.9-3

Security Update?
----------------

No

Build Order
-----------

```
#buildit linux-kernel
```

Test Build(s) Done
------------------

**Primary Architectures**

### DIFF
--- a/runtime-kernel/linux-kernel/autobuild/patches/0329-cpufreq-loongson3-Support-for-older-SMC-firmware-ABI.patch
+++ b/runtime-kernel/linux-kernel/autobuild/patches/0329-cpufreq-loongson3-Support-for-older-SMC-firmware-ABI.patch
@@ -1,0 +1,533 @@
+From feec9a56c7936573766d1713cdc8c57a77f30918 Mon Sep 17 00:00:00 2001
+From: Yao Zi <ziyao@disroot.org>
+Date: Mon, 2 Jun 2025 11:54:17 +0000
+Subject: [PATCH] cpufreq: loongson3: Support for older SMC firmware ABI
+
+SMC firmware found on many on-market LoongArch devices implement a
+different ABI than what has been implemented in the current upstream
+driver. They significantly differ in the following ways:
+
+- CMD_GET_VERSION returns 0.
+- There is no known SMC call to obtain corresponding frequencies for
+  each frequency level. The frequency table must therefore be calculated
+  with CPU clock frequency from scratch.
+- There is no known SMC call to obtain the current frequency level.
+- The host must determine the set of cores able to run at boost
+  frequency and enable boosting manually.
+- SMC call reponse format may vary between commands.
+
+This patch adds support for the SMC firmware ABI found on currently
+available devices, which I denoted as "ABI-1" in the driver. Boost
+support is omitted for now, since determination of cores able to boost
+requires the driver to couple tightly with micro-architecture details.
+
+For coexistence of the two ABIs, I prefixed all SMC-call constants with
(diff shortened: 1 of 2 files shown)


xry111:
I really don't like the terminology "ABIv1."

Those pairs are in fact, completely different and orthogonal concepts:

- Downstream "Linux 4.19.0" UAPI vs. Linux 5.19.0 UAPI
- Stack-based relocs vs. relocs directly operating on the immediate slot
- BPI vs. UEFI
- SMC shipped with current HW vs. SMC implementing the interface expected by the upstream kernel
- Maybe others...

Mixing up those concepts into "ABIv1" vs "ABIv2" *has already* caused many misunderstandings.

ziyao233:
On Fri, Jun 06, 2025 at 03:55:20AM -0700, Xi Ruoyao wrote:
> xry111 left a comment (AOSC-Dev/aosc-os-abbs#11270)
> 
> I really don't like the terminology "ABIv1."
> 
> Those pairs are in fact, completely different and orthogonal concepts:
> 
> - Downstream "Linux 4.19.0" UAPI vs. Linux 5.19.0 UAPI
> - Stack-based relocs vs. relocs directly operating on the immediate slot
> - BPI vs. UEFI
> - SMC shipped with current HW vs. SMC implementing the interface expected by the upstream kernel
> - Maybe others...
> 
> Mixing up those concepts into "ABIv1" vs "ABIv2" *has already* caused many misunderstandings.

Yeah, I've considered the risk of mixing things up as well, but didn't
come up with a reasonable name. So I just stopped waiting and made up a
quick patch with a draft commit message for testing.

As suggested by @xen0n, maybe SMCv0 is a better name, since the easist
way to distinguish these two versions of firmware is by return value of
CMD_GET_VERSION, where the older returns 0.

ziyao233:
@KatyushaScarlet reported that the CPUFreq driver causes frequently amdgpu resets on TongFang L820-T2 when using amdgpu instead of radeon driver. I'll dig it further later.

KatyushaScarlet:
> @KatyushaScarlet reported that the CPUFreq driver causes frequently amdgpu resets on TongFang L820-T2 when using amdgpu instead of radeon driver. I'll dig it further later.


1. Log 1: After unlocking the screen in KDE6 and clicking the tray icon, the desktop froze, blinked, and reported a GPU reset.
[dmesg-l860-t2-amdgpu-reset-6.14.9-log1.log](https://github.com/user-attachments/files/20636981/dmesg-l860-t2-amdgpu-reset-6.14.9-log1.log)


2. Log 2: After reproducing the same steps, the GPU reset occurred. I then removed the loongson3_cpufreq driver using `rmmod`, and the GPU stopped resetting. However, when I reinserted the driver using `modprobe`, the GPU reset again.
[dmesg-l860-t2-amdgpu-reset-6.14.9-log2.log](https://github.com/user-attachments/files/20636982/dmesg-l860-t2-amdgpu-reset-6.14.9-log2.log)

xry111:
Note that the serial debug output (if enabled) from the "old" SMC already announces itself as "v1.x" :(.

PR #11595, opened 2025-07-04
verilator: fix data install path

Mikura-Kyouka:
Topic Description
-----------------
- verilator: fix data install path
    - The original patch 0001 defines `pkglibdir` to changes the install path of binary, for architecture relevant files need to be placed at `/usr/lib` instead of `usr/share` .
    - But verilator needs some other files such as `verilated.h` during compiling verilog. While compiling, it tries finding these files in `pkglibdir`, but they are not installed to there. 
    - The newly added patch 0002 makes files needed by verilator installed to `pkglibdir` , too.

Package(s) Affected
-------------------
- verilator: 5.026

Security Update?
----------------
No

Build Order
-----------

```
#buildit verilator
```

Test Build(s) Done
------------------

**Primary Architectures**

- [ ] AMD64 `amd64`
- [ ] AArch64 `arm64`

**Secondary Architectures**

- [ ] Loongson 3 `loongson3`
- [ ] RISC-V 64-bit `riscv64`

### DIFF
--- a/app-electronics/verilator/autobuild/patches/0002-Fix-data-install-path.patch
+++ b/app-electronics/verilator/autobuild/patches/0002-Fix-data-install-path.patch
@@ -1,0 +1,55 @@
+From af547ed4fb2ae981ccd860fd669aa5f987b96e9b Mon Sep 17 00:00:00 2001
+From: DehaoGao <13207280@qq.com>
+Date: Fri, 4 Jul 2025 14:02:39 +0800
+Subject: [PATCH] Fix data install path
+
+---
+ Makefile.in | 20 ++++++++++----------
+ 1 file changed, 10 insertions(+), 10 deletions(-)
+
+diff --git a/Makefile.in b/Makefile.in
+index d963d8402..c4741d492 100644
+--- a/Makefile.in
++++ b/Makefile.in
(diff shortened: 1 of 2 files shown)


app-electronics/verilator/autobuild/patches/0002-Fix-data-install-path.patch
```diff
@@ -0,0 +1,55 @@
+From af547ed4fb2ae981ccd860fd669aa5f987b96e9b Mon Sep 17 00:00:00 2001
```
xtexx, line 1: Please add some explanations why this addresses the problem? `pkglibdir` is not defined in the source.

Mikura-Kyouka: The patch 0001 only changes the install path of binary, and verilator needs some other files such as `verilated.h` during compiling verilog. Then verilator tries finding these files in `pkglibdir` , but they are not installed to there. `pkglibdir` is defined in patch 0001.

xtexx: ah

MingcongBai (approved):

PR #11694, opened 2025-07-11
box64: new,0.3.6

grayawa:
Topic Description
-----------------

- box64: new, 0.3.6

Package(s) Affected
-------------------

- box64: 0.3.6

Security Update?
----------------

No

Build Order
-----------

```
#buildit box64
```

Test Build(s) Done
------------------

**Primary Architectures**

- [ ] AMD64 `amd64`
- [x] AArch64 `arm64`

**Secondary Architectures**

- [ ] Loongson 3 `loongson3`
- [x] RISC-V 64-bit `riscv64`

### DIFF
--- a/app-emulation/box64/autobuild/alternatives
+++ b/app-emulation/box64/autobuild/alternatives
@@ -1,0 +1,1 @@
+alternative /usr/lib/binfmt.d/emu-x86_64.conf /usr/lib/binfmt.alternatives/box64.conf 70
--- a/app-emulation/box64/autobuild/beyond
+++ b/app-emulation/box64/autobuild/beyond
@@ -1,0 +1,6 @@
+abinfo "Setting executable bits on shared objects ..."
+chmod -v +x "$PKGDIR/usr/lib/box64-x86_64-linux-gnu/"*
+abinfo "Moving binfmt configuration to /usr/lib/binfmt.alternatives ..."
+mkdir -v "$PKGDIR/usr/lib/binfmt.alternatives/"
+mv -v "$PKGDIR/etc/binfmt.d/box64.conf" "$PKGDIR/usr/lib/binfmt.alternatives/"
+rm -rv "$PKGDIR/etc/binfmt.d/"
--- a/app-emulation/box64/autobuild/defines
+++ b/app-emulation/box64/autobuild/defines
@@ -1,0 +1,37 @@
+PKGNAME=box64
+PKGSEC=utils
+PKGDEP="glibc gcc-runtime"
+BUILDDEP="mold"
+PKGDES="Userspace emulator for running x86-64 applications on non-x86 systems"
+ABTYPE="cmakeninja"
+
+# Note: -DCI=ON, regenerates function wrappers, which is unnecessary for packaging?
+CMAKE_AFTER=(
+    "-DCI=ON"
+    "-DWITH_MOLD=ON"
+)
+
+CMAKE_AFTER__ARM64=(
+    "${CMAKE_AFTER[@]}"
+    "-DARM64=ON"
+    "-DARM_DYNAREC=ON"
+)
+
+CMAKE_AFTER__LOONGARCH64=(
+    "${CMAKE_AFTER[@]}"
+    "-DLARCH64=ON"
+    "-DLARCH64_DYNAREC=ON"
+)
+
+CMAKE_AFTER__PPC64EL=(
+    "${CMAKE_AFTER[@]}"
+    "-DPPC64LE=ON"
+)
+
+CMAKE_AFTER__RISCV64=(
+    "${CMAKE_AFTER[@]}"
+    "-DRV64=ON"
+    "-DRV64_DYNAREC=ON"
+)
+
+FAIL_ARCH="!(arm64|loongarch64|ppc64el|riscv64)"
--- a/app-emulation/box64/spec
+++ b/app-emulation/box64/spec
@@ -1,0 +1,4 @@
+VER=0.3.6
+SRCS="git::commit=tags/v$VER::https://github.com/ptitSeb/box64.git"
+CHKSUMS="SKIP"
+CHKUPDATE="anitya::368953"


app-emulation/box64/autobuild/beyond
```diff
@@ -0,0 +1,3 @@
+abinfo "making libs executable..."
+chmod -v +x ""$PKGDIR/usr/lib/box64-x86_64-linux-gnu/*""
+abinfo "all done!"
```
chenx97: Why write this?

app-emulation/box64/autobuild/defines
```diff
@@ -0,0 +1,9 @@
+PKGNAME=box64
+PKGSEC=utils
+PKGDES="an userspace x86_64 emulator with a twist, aimed to run x86_64 Linux programs on non-x86_64 Linux systems"
+PKGDEP="glibc gcc gcc-runtime"
+BUILDDEP="cmake python-3"
+FAIL_ARCH="(amd64,loongson3)"
+CMAKE_AFTER__PPC64EL="-D PPC64LE=1"
```
chenx97:
```bash
CMAKE_AFTER__PPC64EL=("-DPPC64LE=ON")
```
Same issue for loongarch64 and riscv64.

chenx97 (changes requested):

app-emulation/box64/autobuild/beyond
grayawa: if not making the libs executable,autobuild4 would say
[ERROR]: QA (E324): non-executable shared object(s) found in /usr/lib:

app-emulation/box64/autobuild/defines
grayawa: The box64 build guide said the build parameter is gonna to like this...
see https://github.com/ptitSeb/box64/blob/main/docs/COMPILE.md#for-powerpc-64-le

grayawa: fixed.

app-emulation/box64/autobuild/beyond
```diff
@@ -0,0 +1,4 @@
+abinfo "making libs executable..."
+chmod -v +x "$PKGDIR/usr/lib/box64-x86_64-linux-gnu/"*
+abinfo "removing default box64.conf..."
```
MingcongBai: Why not just move this to `/usr/lib/binfmt.d`?

app-emulation/box64/autobuild/overrides/usr/lib/binfmt.alternatives/box64.conf
```diff
@@ -0,0 +1 @@
+:box64:M::\x7f\x45\x4c\x46\x02\x01\x01\x00\x00\x00\x00\x00\x00\x00\x00\x00\x02\x00\x3e\x00:\xff\xff\xff\xff\xff\xfe\xfe\xfc\xff\xff\xff\xff\xff\xff\xff\xff\xfe\xff\xff\xff:/usr/bin/box64:CF
```
MingcongBai: This makes no sense and differs from the upstream copy.

Overriding upstream-supplied files is asking for trouble - you won't notice upstream changes this way.

app-emulation/box64/autobuild/defines
```diff
@@ -0,0 +1,31 @@
...
+    "-DCI=ON"
+    "-DWITH_MOLD=ON"
+)
+CMAKE_AFTER__RISCV64=(
+    "-DRV64=ON"
+    "-DRV64_DYNAREC=ON"
+    "-DCI=ON"
+    "-DWITH_MOLD=ON"
+)
+CMAKE_AFTER__ARM64=("-DARM64=ON"
```
MingcongBai: Can you at least check your code styling?

Also, `CMAKE_AFTER__*` is meant as an overlay:

```
CMAKE_AFTER=(
    '-DCI=ON'
    '-DWITH_MOLD=ON'
)
CMAKE_AFTER__ARM64=(
    "${CMAKE_AFTER[@]}"
    '-DARM64=ON'
    '-DARM_DYNAREC=ON'
)
...
```

Also, at least sort the variables by name so it's easier to find?

app-emulation/box64/autobuild/defines
```diff
@@ -0,0 +1,31 @@
+PKGNAME=box64
+PKGSEC=utils
+PKGDES="an userspace x86_64 emulator with a twist, aimed to run x86_64 Linux programs on non-x86_64 Linux systems"
```
MingcongBai: This is quite strictly against the Styling Manual:

https://wiki.aosc.io/zh/developer/packaging/package-styling-manual/#ruan-jian-bao-shuo-ming

- You did not capitalise the first letter.
- You reproduced upstream propaganda (WTF was that "twist"?) - think: does it help the user?
- Have you noticed that people are using Box64 with Wine, too?
- Does Box64 work on x86 systems? I think not.
- Terminlogy - it's x86-64.

How about:

```
Userspace emulator for running x86-64 applications on non-x86 systems
```

app-emulation/box64/autobuild/beyond
```diff
@@ -0,0 +1,4 @@
+abinfo "making libs executable..."
+chmod -v +x "$PKGDIR/usr/lib/box64-x86_64-linux-gnu/"*
+abinfo "moving default box64.conf to /usr/lib/binfmt.alternatives..."
+mv "$PKGDIR/etc/binfmt.d/box64.conf" "/usr/lib/binfmt.alternatives/"
```
MingcongBai: Use verbosity `mv -v`.

MingcongBai: That directory does not exist - this is bound to fail.

MingcongBai: And you would need to remove `"$PKGDIR"/etc/binfmt.d`.

app-emulation/box64/autobuild/beyond
```diff
@@ -0,0 +1,4 @@
+abinfo "making libs executable..."
+chmod -v +x "$PKGDIR/usr/lib/box64-x86_64-linux-gnu/"*
+abinfo "moving default box64.conf to /usr/lib/binfmt.alternatives..."
+mv -v "$PKGDIR/etc/binfmt.d/box64.conf" "$PKGDIR/usr/lib/binfmt.alternatives/"
```
MingcongBai, line 5: This is bound to fail - you have not created the target directory yet.

Also, remove `"$PKGDIR"/etc/binfmt.d` after the move.

app-emulation/box64/autobuild/defines
```diff
@@ -0,0 +1,33 @@
...
+PKGDEP="glibc gcc-runtime"
+ABTYPE="cmakeninja"
+BUILDDEP="mold"
+
+FAIL_ARCH="(amd64|loongson3)"
+
+# No need to regenerate wrapper,-DCI=ON
+CMAKE_AFTER__PPC64EL=(
```
MingcongBai: `CMAKE_AFTER__*` is meant as an overlay:

```bash
CMAKE_AFTER=(
    '-DCI=ON'
    '-DWITH_MOLD=ON'
)
CMAKE_AFTER__ARM64=(
    "${CMAKE_AFTER[@]}"
    '-DARM64=ON'
    '-DARM_DYNAREC=ON'
)
...
```

Also, at least sort the variables by name so they are easier to find?

app-emulation/box64/autobuild/defines
```diff
@@ -0,0 +1,33 @@
+PKGNAME=box64
+PKGSEC=utils
+PKGDES="an userspace x86_64 emulator with a twist, aimed to run x86_64 Linux programs on non-x86_64 Linux systems"
+PKGDEP="glibc gcc-runtime"
+ABTYPE="cmakeninja"
+BUILDDEP="mold"
+
+FAIL_ARCH="(amd64|loongson3)"
+
+# No need to regenerate wrapper,-DCI=ON
```
MingcongBai:
```
# Note: -DCI=ON, regenerates function wrappers, which is unnecessary for packaging?
```

If I were you, though, I would not press a button that I do not know the full consequence to - and so I wouldn't with a random CMake variable.

If the default `-DCI=OFF` didn't cause any issue, don't change it.

app-emulation/box64/autobuild/defines
```diff
@@ -0,0 +1,33 @@
+PKGNAME=box64
+PKGSEC=utils
+PKGDES="an userspace x86_64 emulator with a twist, aimed to run x86_64 Linux programs on non-x86_64 Linux systems"
+PKGDEP="glibc gcc-runtime"
+ABTYPE="cmakeninja"
+BUILDDEP="mold"
+
+FAIL_ARCH="(amd64|loongson3)"
```
MingcongBai: This is a false assumption - we have many more architectures than just the six we have included for the mainline so far. If the fact is that "only four architectures are supported", your FAIL_ARCH should then say "if the architecture is not found amongst the four, then fail":

```
FAIL_ARCH="!(arm64|loongarch64|ppc64el|riscv64)"
```

app-emulation/box64/autobuild/defines
```diff
@@ -0,0 +1,33 @@
+PKGNAME=box64
+PKGSEC=utils
+PKGDES="an userspace x86_64 emulator with a twist, aimed to run x86_64 Linux programs on non-x86_64 Linux systems"
+PKGDEP="glibc gcc-runtime"
+ABTYPE="cmakeninja"
```
MingcongBai: Please move this to below FAIL_ARCH, for readability's sake.

app-emulation/box64/autobuild/defines
```diff
@@ -0,0 +1,33 @@
...
+BUILDDEP="mold"
+
+FAIL_ARCH="(amd64|loongson3)"
+
```
MingcongBai, line 13: Move ABTYPE here.

app-emulation/box64/autobuild/alternatives
```diff
@@ -0,0 +1 @@
+alternative /usr/lib/binfmt.d/emu-x86_64.conf /usr/lib/binfmt.alternatives/box64.conf 100
```
MingcongBai: Really, 100? Are we so confident that it should be the default?

app-emulation/box64/autobuild/beyond
```diff
@@ -0,0 +1,4 @@
+abinfo "making libs executable..."
```
MingcongBai:
```
abinfo "Setting executable bits on shared objects ..."
```

Static libraries are "libs" too, but they should not be executable.

app-emulation/box64/autobuild/beyond
```diff
@@ -0,0 +1,4 @@
...
+chmod -v +x "$PKGDIR/usr/lib/box64-x86_64-linux-gnu/"*
+abinfo "moving default box64.conf to /usr/lib/binfmt.alternatives..."
```
MingcongBai:
```
abinfo "Moving binfmt configuration to /usr/lib/binfmt.alternatives ..."
```

MingcongBai (changes requested):

app-emulation/box64/autobuild/alternatives
grayawa: Yes.QEMU is default now,and it's not working,in my test.
Box64 is working out-of-the-box,so I think it should be the default.

app-emulation/box64/autobuild/beyond
grayawa: they are libs,but if it's not been made executable,abbs will yell.

app-emulation/box64/autobuild/beyond
grayawa: solved,I think.

app-emulation/box64/autobuild/defines
grayawa: I think I've done it.

app-emulation/box64/autobuild/alternatives
grayawa: whatever,changed

grayawa:
tested on arm64 
image
the lack of lib error should be fine...

MingcongBai (approved):

chenx97 (approved):

chenx97 (approved):

PR #11866, opened 2025-07-26
nginx: add ngx_brotli module

abc1763613206:
Topic Description
-----------------

- nginx: add ngx_brotli module

Package(s) Affected
-------------------

- nginx: 1.28.0-1

Security Update?
----------------

No

Build Order
-----------

```
#buildit nginx
```

Test Build(s) Done
------------------

**Primary Architectures**

- [x] AMD64 `amd64`
- [x] AArch64 `arm64`

**Secondary Architectures**

- [x] Loongson 3 `loongson3`
- [x] RISC-V 64-bit `riscv64`

### DIFF
--- a/app-web/nginx/autobuild/build
+++ b/app-web/nginx/autobuild/build
@@ -41,7 +41,8 @@ abinfo "Configuring nginx ..."
     --with-stream_realip_module \
     --with-stream_ssl_module \
     --with-stream_ssl_preread_module \
-    --with-threads
+    --with-threads \
+    --add-module="$SRCDIR/../ngx_brotli"
 
 abinfo "Building nginx ..."
 make
--- a/app-web/nginx/spec
+++ b/app-web/nginx/spec
@@ -1,5 +1,12 @@
-# Stable branch
-VER=1.28.0
-SRCS="tbl::https://nginx.org/download/nginx-$VER.tar.gz"
-CHKSUMS="sha256::c6b5c6b086c0df9d3ca3ff5e084c1d0ef909e6038279c71c1c3e985f576ff76a"
+NGINX_VER=1.28.0
+BROTLI_VER=1.0.0~rc1+git20231009
+VER="${NGINX_VER}+brotli${BROTLI_VER}"
+# Note: After v1.0.0rc, there are still some necessary updates (including upstream submodules), 
+# so the following commit refers to the latest one in master branch.
+SRCS="tbl::https://nginx.org/download/nginx-$NGINX_VER.tar.gz \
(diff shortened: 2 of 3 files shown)


abc1763613206:
image

abc1763613206:
Add the following configuration to `nginx.conf`:
image
Then use curl to specify the request headers for testing, and **it works** smoothly:
image

app-web/nginx/autobuild/build
```diff
@@ -41,7 +41,8 @@ abinfo "Configuring nginx ..."
     --with-stream_realip_module \
     --with-stream_ssl_module \
     --with-stream_ssl_preread_module \
-    --with-threads
+    --with-threads \
+    --add-module=$SRCDIR/../ngx_brotli
```
MingcongBai: Please quote `$SRCDIR`.

app-web/nginx/spec
```diff
@@ -1,5 +1,9 @@
 # Stable branch
 VER=1.28.0
-SRCS="tbl::https://nginx.org/download/nginx-$VER.tar.gz"
-CHKSUMS="sha256::c6b5c6b086c0df9d3ca3ff5e084c1d0ef909e6038279c71c1c3e985f576ff76a"
+SRCS="tbl::https://nginx.org/download/nginx-$VER.tar.gz \
+      git::commit=a71f9312c2deb28875acc7bacfdd5695a111aa53;rename=ngx_brotli::https://github.com/google/ngx_brotli.git"
```
MingcongBai, line 7: Why not use a tag?

app-web/nginx/spec
```diff
@@ -1,5 +1,9 @@
...
+CHKSUMS="sha256::c6b5c6b086c0df9d3ca3ff5e084c1d0ef909e6038279c71c1c3e985f576ff76a \
+         SKIP"
 CHKUPDATE="anitya::id=5413"
+SUBDIR="nginx-$VER"
+REL=1
```
MingcongBai: Instead of bumping REL, you should be using a compound version, something like:

```
1.28.0+ngx_brotli1.0.0~rc1
```

So that all components of this compound source is represented.

MingcongBai (changes requested):

app-web/nginx/spec
abc1763613206: Because the upstream made some additional changes after v1.0.0rc, including submodule synchronization with brotli updates, all in the master branch. This is actually the latest commit in master.

hanyuwei70: If so, I think it's better in topic.

MingcongBai (approved):

PR #12356, opened 2025-09-02
js-102: add loongarch64_nosimd defines

AirFortressIlikara:
Topic Description
-----------------

- js-102: add loongarch64_nosimd defines

Package(s) Affected
-------------------

- js-102: 102.15.1

Security Update?
----------------

No

Build Order
-----------

```
#buildit js-102
```

Test Build(s) Done
------------------

**Primary Architectures**

- [ ] AMD64 `amd64`
- [ ] AArch64 `arm64`

**Secondary Architectures**

- [ ] Loongson 3 `loongson3`
- [ ] RISC-V 64-bit `riscv64`

### DIFF
--- a/lang-js/js-102/autobuild/defines
+++ b/lang-js/js-102/autobuild/defines
@@ -22,6 +22,8 @@ AUTOTOOLS_AFTER="--disable-debug
 AUTOTOOLS_AFTER__LOONGARCH64=" \
                  ${AUTOTOOLS_AFTER} \
                  --enable-linker=bfd"
+AUTOTOOLS_AFTER__LOONGARCH64_NOSIMD=" \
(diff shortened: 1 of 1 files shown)


xry111:
Ok but is there any significant necessity to run GNOME on 2k0300?

xry111 (dismissed):
Please increase PKGREL.

MingcongBai (approved):
No need to bump REL for a new architecture.

xry111:
> No need to bump REL for a new architecture.

FTR I still think it's needed.

If we don't do that, the same REL on different architectures will be built with different versions of the toolchain components.  In this case, at least different GCC and Binutils versions.

Then if a behavior difference is detected on different architectures, the same REL may puzzle us when we diagnostic it.

To me the technical status is already different enough for us to distinguish with different REL.

PR #12373, opened 2025-09-03
aosc-os-presets-desktop: Don't enable avahi-daemon.service by default

xry111:
Topic Description
-----------------

systemd-resolved already has the mDNS transponder functionality, and it's enabled by default.  Thus having systemd-resolved running together with avahi-daemon will cause a conflict.

The conflict is a contribute factor of the very annoying "avahi-daemon keeps reporting a conflict and bump the ID in the host name" issue, but that issue is not solely caused by the conflict (i.e. it even happens without the conflict).  Considering this issue is still not fixed (nor even understood) after years it was reported, to me we should just keep using the systemd mDNS transponder and disable avahi by default.

If a user must use avahi they need to at least set MulticastDNS=no in systemd/resolved.conf.d/ and if they still hit the annoying issue I guess we've no way to help them.

Package(s) Affected
-------------------

aosc-os-presets-desktop: `6`

Security Update?
----------------

No

Build Order
-----------

```
#buildit aosc-os-presets-desktop
```

Test Build(s) Done
------------------

**Primary Architectures**

- [x] AMD64 `amd64`
- [x] AArch64 `arm64`

**Secondary Architectures**

- [x] Loongson 3 `loongson3`
- [x] RISC-V 64-bit `riscv64`

### DIFF
--- a/runtime-data/aosc-os-presets-desktop/autobuild/build
+++ b/runtime-data/aosc-os-presets-desktop/autobuild/build
@@ -17,7 +17,6 @@ enable accounts-daemon.service
 
 # Network services
 enable smbd.service
-enable avahi-daemon.service
 enable avahi-dnsconfd.service
 EOF
 
--- a/runtime-data/aosc-os-presets-desktop/spec
+++ b/runtime-data/aosc-os-presets-desktop/spec
@@ -1,2 +1,2 @@
-VER=5
+VER=6
 DUMMYSRC=1


xry111:
References:
- https://github.com/avahi/avahi/issues/117
- https://github.com/avahi/avahi/issues/117#issuecomment-2666840702
- https://wiki.archlinux.org/index.php/Avahi says: "[systemd-resolved](https://wiki.archlinux.org/title/Systemd-resolved#mDNS) has a built-in mDNS service, make sure to disable systemd-resolved's multicast DNS resolver/responder (refer to [resolved.conf(5)](https://man.archlinux.org/man/resolved.conf.5)) or [disable](https://wiki.archlinux.org/title/Disable) systemd-resolved.service entirely before using Avahi."

MingcongBai (approved):

PR #12477, opened 2025-09-10
fex: new, 2509

HmnSn:
Topic Description
-----------------

- fex: new, 2509

Package(s) Affected
-------------------

- fex: 2509

Security Update?
----------------

No

Build Order
-----------

```
#buildit fex
```

Test Build(s) Done
------------------

**Primary Architectures**

- [x] AArch64 `arm64`

### DIFF
--- a/app-emulation/fex/autobuild/defines
+++ b/app-emulation/fex/autobuild/defines
@@ -1,0 +1,21 @@
+PKGNAME=fex
+PKGSEC=utils
+PKGDEP="gcc-runtime glibc qt-6 xxhash curl \
+        squashfuse squashfs-tools erofs-utils"
+# Note:
+# FEXRootFSFetcher can fall back to TTY when zenity is not available.
+PKGRECOM="zenity"
+BUILDDEP="llvm"
+PKGDES="Usermode x86 and x86-64 emulator for ARM64"
+
+USECLANG=1
+
+ABTYPE=cmakeninja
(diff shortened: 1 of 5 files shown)


HmnSn:
Tested good with Linux/AMD64 Geekbench5 and Wine + CPU-Z.

https://browser.geekbench.com/v5/cpu/23776807

Screenshot_20250911_104259
Screenshot_20250911_104344

app-emulation/fex/autobuild/defines
```diff
@@ -0,0 +1,19 @@
+PKGNAME=fex
+PKGSEC=utils
+PKGDEP="gcc-runtime glibc qt-6 xxhash curl \
+        squashfuse squashfs-tools erofs-utils"
+PKGRECOM="zenity"
```
MingcongBai, line 7: Why was this a recommends? Would it be required at any point?

MingcongBai (changes requested):

app-emulation/fex/autobuild/defines
HmnSn: Because FEXRootFSFetcher can fall back to TTY when zenity is not available.

MingcongBai: Understood, can you add this as a comment as it's not immediately clear?

MingcongBai (approved):

PR #12559, opened 2025-09-16
systemd: PKGPROV += udev

xry111:
Topic Description
-------------------

Make "udev" an alias of systemd.

Package(s) Affected
-------------------

- systemd: `257.6-2`

Security Update?
----------------

No

Build Order
-----------

```
#buildit systemd
```

Test Build(s) Done
------------------

**Primary Architectures**

- [x] AMD64 `amd64`
- [x] AArch64 `arm64`

**Secondary Architectures**

- [x] Loongson 3 `loongson3`
- [x] RISC-V 64-bit `riscv64`

### DIFF
--- a/app-admin/systemd/autobuild/defines
+++ b/app-admin/systemd/autobuild/defines
@@ -287,4 +287,4 @@ PKGEPOCH=1
 PKGBREAK="openresolv<=3.9.0 \
           dracut<=059-4" # Missing systemd-executor, cannot boot
 PKGREP="openresolv<=3.9.0"
-PKGPROV="openresolv==1:0"
+PKGPROV="openresolv==1:0 udev_spiral"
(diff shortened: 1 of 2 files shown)


app-admin/systemd/autobuild/defines
```diff
@@ -287,4 +287,4 @@ PKGEPOCH=1
 PKGBREAK="openresolv<=3.9.0 \
           dracut<=059-4" # Missing systemd-executor, cannot boot
 PKGREP="openresolv<=3.9.0"
-PKGPROV="openresolv==1:0"
+PKGPROV="openresolv==1:0 udev"
```
MingcongBai: This should be udev_spiral so that it matches it with an exact version ==$PKGVER

MingcongBai (changes requested):

MingcongBai (approved):

PR #12893, opened 2025-10-06
ss-tproxy: revamp postinst and spec

chenx97:
Topic Description
-----------------

- ss-tproxy: revamp postinst and spec
    - use git-based SRCS\[^1\]
    - properly treat "exist" as a verb
    - reload the systemd configuration
    - use \`is-active\` in a more elegant manner
    - with \`--quiet\` we just need to detect if systemctl returns 0
    - use \`try-restart\` instead
    - allow restarts to fail as we should allow users to debug their
    own configurations later
    \[^1\]: https://github.blog/open-source/git/update-on-the-future-stability-of-source-code-archives-and-hashes/#future-stability-of-archives-and-hashes

Package(s) Affected
-------------------

- ss-tproxy: 4.8.3-1

Security Update?
----------------

No

Build Order
-----------

```
#buildit ss-tproxy
```

Test Build(s) Done
------------------

**Primary Architectures**

### DIFF
--- a/app-proxy/ss-tproxy/autobuild/postinst
+++ b/app-proxy/ss-tproxy/autobuild/postinst
@@ -1,8 +1,13 @@
+echo "Refreshing systemd configurations ..."
+systemctl daemon-reload
 
-echo "Check whether ss-tproxy is exist..."
-if [ $(systemctl is-enabled ss-tproxy) == "enabled" ]; then
- echo "Starting ss-tproxy service..."
- systemctl start ss-tproxy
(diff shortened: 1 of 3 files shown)


chenx97:
fix pending: try-restart is a no-op when the unit is not active.

chenx97:
image

Upgrading works

image

Reinstalling with disabled and inactive unit works

image

Reinstalling with active unit works

image

Reinstalling with enabled but inactive unit works

image

Installation doesn't fail when the systemd unit fails

MingcongBai (approved):

PR #13242, opened 2025-10-25
fontconfig: remove unconditional call to systemctl enable

MingcongBai:
Topic Description
-----------------

- aosc-os-presets-desktop: enable fc-cache.service
- fontconfig: remove unconditional call to systemctl enable
    This should be handled in preset \(and will be with
    aosc-os-presets-desktop\). If a service were to be enabled in postinst and
    such an \`systemctl enable' call were to be made via a dpkg trigger during
    OOBE/firstboot, it causes a call to \`systemctl daemon-reload', which would
    terminate oneshot services such as our OOBE service as it times out after
    the reload.
    Remove this call from postinst to prevent this issue \(and that it would be
    more appropriate\).

Package(s) Affected
-------------------

- aosc-os-presets-desktop: 8
- fontconfig: 1:2.15.0-2

Security Update?
----------------

No

Build Order
-----------

```
#buildit fontconfig aosc-os-presets-desktop
```

Test Build(s) Done
------------------

**Primary Architectures**

- [x] AMD64 `amd64`
- [x] AArch64 `arm64`

**Secondary Architectures**

- [x] Loongson 3 `loongson3`
- [x] RISC-V 64-bit `riscv64`

### DIFF
--- a/runtime-data/aosc-os-presets-desktop/autobuild/build
+++ b/runtime-data/aosc-os-presets-desktop/autobuild/build
@@ -25,6 +25,9 @@ if ab_match_arch loongarch64; then
 
 # LoongGPU OpenGL runtime switching service
 enable loonggl-sw.service
+
+# Fontconfig cache generation
+enable fc-cache.service
 EOF
 fi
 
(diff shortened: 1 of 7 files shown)


Cyanoxygen (approved):
`fontconfig` trigger no longer triggers `daemon-reload` of systemd. Yippie!

PR #14121, opened 2025-12-19
Core 13.0.1

MingcongBai:
Topic Description
-----------------

- topics: add core-13.0.1.toml
- aosc-aaa: update to 13.0.1
- glibc: \(loongson3\) allow dlopen shared objects with executable stack for now
    From the commit msg:
    Commit 0ca8785a2851 \("elf: Do not change stack permission on
    dlopen/dlmopen"\) has disabled dlopen shared objects requiring executable
    stack by default unless the program or one of its dependency \(via
    DT_NEEDED, either direct or indirect\) already requires executable stack.
    Unfortunately, on loongson3 GCC had had a bug causing many shared
    objects incorrectly announcing the requirement of executable stack.  The
    bug was only "recently" fixed at https://gcc.gnu.org/r15-3205.  So if
    the program and all of its dependencies are built with Core 13, but the
    dlopen'ed shared object is built with Core 12, the dlopen will fail.
    This seems causing a massive system breakage on loongson3.
    Change the default for now until we rebuild everything with Core 13.

Package(s) Affected
-------------------

- aosc-aaa: 13.0.1
- glibc: 2.42-1

Security Update?
----------------

No

Build Order
-----------

```
#buildit glibc aosc-aaa
```

Test Build(s) Done
------------------

**Primary Architectures**

- [x] AMD64 `amd64`
- [x] AArch64 `arm64`

**Secondary Architectures**

- [ ] Loongson 3 `loongson3`
- [ ] RISC-V 64-bit `riscv64`


MingcongBai:
TUM:

image

MingcongBai:
Before:

image

After:

image

chenx97 (approved):

PR #14044, opened 2025-12-15
OpenCV 4.12 Survey

stydxm:
Topic Description
-----------------

- opencv: update to 4.12.0

Package(s) Affected
-------------------

- opencv: 4.12.0

Security Update?
----------------

No

Build Order
-----------

```
#buildit opencv:-pkgbreak gavl digikam frei0r-plugins gmic gstreamer mlt openimageio spectacle opencv
```

Test Build(s) Done
------------------

**Primary Architectures**

- [x] AMD64 `amd64`
- [x] AArch64 `arm64`
- [x] LoongArch 64-bit `loongarch64`
- [x] LoongArch 64-bit (No SIMD) `loongarch64_nosimd`

**Secondary Architectures**

- [x] Loongson 3 `loongson3`
- [x] PowerPC 64-bit (Little Endian) `ppc64el`
- [x] RISC-V 64-bit `riscv64`

PR #14148, opened 2025-12-21
extra-cmake-modules: Disable --fatal-warnings in linker flags

Cyanoxygen:
Topic Description
-----------------

- extra-cmake-modules: disable --fatal-warnings for linker
    We now have GCS warnings when linking to shared objects that aren't
    built with GCS \(this is only being default since GCC 15\).

Package(s) Affected
-------------------

- extra-cmake-modules: 5.115.0-1

Security Update?
----------------

No

Build Order
-----------

```
#buildit extra-cmake-modules
```

Test Build(s) Done
------------------

**Primary Architectures**


stydxm (approved):

Cyanoxygen:
Tested with topic #14044.

PR #14354, opened 2026-01-05
loonggl: add postinst to remove /etc/init.d/vga_check.sh

xry111:
Topic Description
-----------------

This file is removed in 1.0.2 but it's treated as "conffile" so dpkg won't remove it.  The rm command is already in prerm but it won't be run if the user updates directly from 1.0.1 to 1.0.2.

Package(s) Affected
-------------------

- loonggl: `1.0.2+lnd25.1~rc1.7-1`

Security Update?
----------------

No

Build Order
-----------

```
#buildit loonggl
```

Test Build(s) Done
------------------

**Primary Architectures**

### DIFF
--- a/runtime-display/loonggl/autobuild/postinst
+++ b/runtime-display/loonggl/autobuild/postinst
@@ -1,0 +1,1 @@
+rm -f /etc/init.d/vga_check.sh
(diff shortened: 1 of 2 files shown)


MingcongBai (changes requested):
Bump REL.

xry111:
🤡

MingcongBai (approved):
LGTM

PR #14418, opened 2026-01-09
loonggl: resurrect the hack previously known as "vga_check.sh"

xry111:
Topic Description
-----------------

But with a better naming, "loonggl_hack."

See the long comment in that script for all the reason of this bullsh*t.

Note that this does NOT fix the WebKitWebProcess crash on LG100/200, it only prevents the crash on platforms where LG100/200 is NOT in use.

Package(s) Affected
-------------------

- loonggl: `1.0.2+lnd25.1~rc1.7-2`

Security Update?
----------------

No

Build Order
-----------

```
#buildit loonggl
```

Test Build(s) Done
------------------

**Primary Architectures**


xry111:
The logic will wrongly treat LoongVPU as a "non-Loong GPU."  Sigh.

Mikura-Kyouka:
LGTM.
图片

Mikura-Kyouka (approved):
LGTM.

PR #14567, opened 2026-01-22
loonggpu-kernel-dkms: add a modprobe.d conf to always load it before loongson

xry111:
Topic Description
-----------------

We used loongson.ls7a2000_support to ensure loongson (upstream DC driver with a very bad name) not to get the way of loonggpu.  But when loonggpu is not installed or fails to work we actually want to use loongson as a fallback.

Add a modprobe.d conf file to ensure loading loonggpu before loongson so we can remove ls7a2000_support later and make the fallback work.  The disadvantage is we cannot use "modprobe.blacklist=loonggpu" anymore: we need to use a custom and not so intuitive "loonggpu.support=0" instead. But overall I still think it will be a win.

Package(s) Affected
-------------------

- loonggpu-kernel-dkms: `1.0.2-lnd25.1~rc1.7-3`

Security Update?
----------------

No

Build Order
-----------

```
#buildit loonggpu-kernel-dkms
```

Test Build(s) Done
------------------

**Primary Architectures**

### DIFF
--- a/runtime-kernel/loonggpu-kernel-dkms/autobuild/overrides/usr/lib/modprobe.d/loonggpu.conf
+++ b/runtime-kernel/loonggpu-kernel-dkms/autobuild/overrides/usr/lib/modprobe.d/loonggpu.conf
@@ -1,0 +1,1 @@
+softdep loongson pre: loonggpu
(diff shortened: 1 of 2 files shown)


MingcongBai (approved):

xry111:
屏幕截图_20260122_171552

屏幕截图_20260122_172522

Tested using vanillarc.

PR #14836, opened 2026-02-18
firefox: fix detection of default browser

chenx97:
Topic Description
-----------------

- firefox: use realpath of the executable for the desktop entry
    Based on code found in \`nsGIOService\` and \`nsGNOMEShellService\`,
    a valid default browser entry for Firefox should have the real path
    to the executable in the Exec field, or Firefox will believe that it
    isn't the default browser. Let's use \`/usr/lib/firefox/firefox\` for
    the Exec field, as that's what the generated entries use.

Package(s) Affected
-------------------

- firefox: 147.0.3-2

Security Update?
----------------

No

Build Order
-----------

```
#buildit firefox
```

Test Build(s) Done
------------------

**Primary Architectures**

- [x] AMD64 `amd64`
- [x] AArch64 `arm64`

**Secondary Architectures**

- [x] Loongson 3 `loongson3`
- [x] RISC-V 64-bit `riscv64`

### DIFF
--- a/app-web/firefox/autobuild/overrides/usr/share/applications/firefox.desktop
+++ b/app-web/firefox/autobuild/overrides/usr/share/applications/firefox.desktop
@@ -42,7 +42,7 @@ Comment[pt]=Navegue na Internet
 Comment[pt_BR]=Navegue na Internet
 Comment[sk]=Prehliadanie internetu
 Comment[sv]=Surfa på webben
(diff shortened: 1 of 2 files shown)


MingcongBai (approved):
Firefox is now set as the default browser out of the box and no longer generates a blank icon on reset.

PR #15020, opened 2026-03-03
python-2: phasing out python-2 dependencies

CyberNeko-AI:
Topic Description
-----------------

- python-is-python3: new
- python-2: fix compatibility with OpenSSL 3 and GCC 15, add python2 update-alternatives
- qt-5: remove python-2 dep, drop qt-5-doc, fix compatibility with python-3.14
- qt-6: remove python-2 dep, fix build with pipewire 1.6, use clang on arm64
- gtk-2+32: remove python-2 dep
- glib+32: remove python-2 dep
- mate-menus: remove python-2 dep
- blastem: remove python-2 dep, fix build with GCC 15

Package(s) Affected
-------------------

- blastem: 0.6.2-3
- glib+32: 2.83.3-1
- mate-menus: 1:1.28.0-1
- python-2: 2.7.18-5
- qt-5: 1:5.15.16+webengine5.15.18+webkit5.212.0+kde20241121-8
- qt-6: 6.10.0-2
- qt-6-doc: 6.10.0-2
- python-is-python3: 1

Security Update?
----------------

No

Build Order
-----------

```
#buildit python-2 blastem mate-menus qt-5 qt-6 python-is-python3
```

Test Build(s) Done
------------------

**Primary Architectures**

- [x] AMD64 `amd64`
- [x] AArch64 `arm64`

**Secondary Architectures**

- [x] Loongson 3 `loongson3`
- [x] RISC-V 64-bit `riscv64`


app-games/blastem/autobuild/build
```diff
@@ -1,6 +1,7 @@
 abinfo "Compiling BlastEm ..."
+
 make \
-    OPT="${CFLAGS} -I/usr/include/SDL2 -lSDL2 -lGLEW -lGL"
+    OPT="${CFLAGS} -I/usr/include/SDL2 -lSDL2 -lGLEW -lGL -Wno-error=incompatible-pointer-types"
```
MingcongBai: You probably wanted `-std=gnu99`, or better yet, patches to modernise their C code:

- Probably this? https://www.retrodev.com/repos/blastem/rev/42c12d141f6e
- This should replace 0001 - https://www.retrodev.com/repos/blastem/rev/e45a317802bd
- Likely also wanted - https://www.retrodev.com/repos/blastem/rev/f973651b48d7

app-games/blastem/autobuild/defines
```diff
@@ -1,7 +1,7 @@
 PKGNAME=blastem
 PKGSEC=games
 PKGDEP="sdl2 glew glu"
-BUILDDEP="make python-2 xcftools pillow vasm"
+BUILDDEP="make xcftools pillow vasm"
 PKGDES="Fast and accurate Genesis emulator"
```
MingcongBai: Emulator for Sega Genesis

groups/libffi-rebuilds
MingcongBai, line 1: Since we are keeping Python 2, drop all changes to groups/.

lang-python/python-2/autobuild/patches/0003-Ubuntu-Fix-OpenSSL-3.6-Compatibility.patch
MingcongBai, line 1: Was this patch derived from an upstream patch? If so, please note.

lang-python/python-2/autobuild/prepare
```diff
@@ -23,3 +23,5 @@ if [[ "${CROSS:-$ARCH}" = "ppc64el" ]]; then
     abinfo "Disabling x86 intrinsics to fix build on ppc64el ..."
     export CPPFLAGS="${CPPFLAGS/-DNO_WARN_X86_INTRINSICS/}"
 fi
+
+export CFLAGS="$CFLAGS -std=gnu11 -Wno-error=incompatible-pointer-types"
```
MingcongBai: Again, see blastem.

runtime-desktop/qt-6/01-runtime/patches/0006-AOSCOS-qtwebengine-fix-build-with-pipewire-1.6.diff.patch
MingcongBai, line 1: Prefix should be UPSTREAM.

MingcongBai: Patch series should be generated with `git format-patch` and this is obviously not it.

runtime-desktop/qt-6/01-runtime/patches/0008-AOSCOS-LOONGARCH64_NOSIMD-disable-LTO.patch.loongarch64_nosimd
```diff
@@ -0,0 +1,18 @@
+Subject: [PATCH 8/8] AOSCOS: LOONGARCH64_NOSIMD: disable LTO to avoid GCC ICE
+
```
MingcongBai: Do we have a fragment of the build log? I seriously doubt (still) that this is specific to this particular subport.

runtime-desktop/qt-6/01-runtime/defines
```diff
@@ -128,3 +128,6 @@ NOSTATIC=0
 
 # Does not work if built with qt 6.6
 PKGBREAK="fcitx5-qt<=1:5.1.3-1 fcitx-qt5<=1.2.7-3"
+
+#FIXME: GCC ICE encountered, fixing by disabling lto.
```
MingcongBai: So are we using the patch or this?

MingcongBai (changes requested):

runtime-desktop/qt-6/01-runtime/defines
CyberNeko-AI: We are using the patch. I've tested and confirmed that adding NOLTO__LOONGARCH64_NOSIMD=1 will not work (because Qt's CMake configuration still ends up mixing `-flto` and `-fno-lto` flags in the same command, which causes a "plugin needed to handle lto object" linker error). 

Using the patch ensures the LTO flags are cleanly and exclusively disabled at the CMake level for this specific architecture. I will remove this leftover line from the `defines` file in the next push.

groups/libffi-rebuilds
CyberNeko-AI: Dropped all changes to groups as requested.

app-games/blastem/autobuild/defines
CyberNeko-AI: Modified PKGDES as requested.

app-games/blastem/autobuild/build
CyberNeko-AI: I've updated the PR with patches to properly modernize the C source code. This includes both upstream fixes and a custom minimal patch, as some upstream refactoring was too massive to backport directly.

Please note that the 0001 patch ([e45a317802bd](https://www.retrodev.com/repos/blastem/rev/e45a317802bd)) actually addresses a different issue than [42c12d141f6e](https://www.retrodev.com/repos/blastem/rev/42c12d141f6e), so it remains necessary.

lang-python/python-2/autobuild/patches/0003-Ubuntu-Fix-OpenSSL-3.6-Compatibility.patch
CyberNeko-AI: Yes, this is actually a consolidated patch. I backported a series of related OpenSSL 3 compatibility patches from Ubuntu, adapted them to our current codebase, and squashed them into this single patch to keep the patch history clean and logical.

I will update the patch header to explicitly list all the original Ubuntu source URLs/commits for proper attribution.

lang-python/python-2/autobuild/prepare
CyberNeko-AI: I've dropped the `-Wno-error=incompatible-pointer-types` workaround and switched to `-std=gnu99`. Since Python 2 is long EOL, it makes the most sense to just build it against the older C standard it was originally designed for.

runtime-desktop/qt-6/01-runtime/patches/0006-AOSCOS-qtwebengine-fix-build-with-pipewire-1.6.diff.patch
CyberNeko-AI: Properly regenerated patches with `git format-patch` and renamed patch 0006.

runtime-desktop/qt-6/01-runtime/patches/0008-AOSCOS-LOONGARCH64_NOSIMD-disable-LTO.patch.loongarch64_nosimd
CyberNeko-AI: I've checked again and confirmed that there's no problem building with lto enabled on `loongarch64_nosimd`. The patch was removed.

runtime-desktop/qt-6/01-runtime/defines
CyberNeko-AI: I've checked again and confirmed that there's no problem building with lto enabled on `loongarch64_nosimd`. These lines were deleted.

CyberNeko-AI:
Removed `glib+32` and `gtk-2+32` from Build Order since they are only available in `amd64`. 
Build of the two packages will be launched manually (as suggested by fellow maintainer).

app-games/blastem/autobuild/build
```diff
@@ -1,4 +1,5 @@
 abinfo "Compiling BlastEm ..."
```
MingcongBai, line 1: Drop this change.

lang-python/python-2/autobuild/defines
```diff
@@ -3,13 +3,16 @@ PKGSEC=python
...
-                 --with-system-ffi"
+AUTOTOOLS_AFTER=(
+    '--enable-shared'
+    '--with-threads'
+    '--enable-ipv6'
+    '--enable-unicode=ucs4'
+    '--with-system-expat'
+    '--with-dbmliborder=gdbm:ndbm'
+    '--with-system-ffi'
+    )
```
MingcongBai: Maybe remove the indentation here.

app-games/blastem/autobuild/build
CyberNeko-AI: Dropped as requested.

lang-python/python-2/autobuild/defines
CyberNeko-AI: Corrected indentation as requested.

app-games/blastem/autobuild/build
```diff
@@ -1,4 +1,3 @@
-abinfo "Compiling BlastEm ..."
```
MingcongBai, line 1: Hmm?

lang-python/python-2/autobuild/prepare
```diff
@@ -23,3 +23,5 @@ if [[ "${CROSS:-$ARCH}" = "ppc64el" ]]; then
     abinfo "Disabling x86 intrinsics to fix build on ppc64el ..."
     export CPPFLAGS="${CPPFLAGS/-DNO_WARN_X86_INTRINSICS/}"
 fi
+
+export CFLAGS="$CFLAGS -std=gnu99"
```
MingcongBai, line 38: Are you sure if this is still needed? If so, please comment above this line a snippet of the build error when built without this flag.

lang-python/python-is-python3/autobuild/defines
```diff
@@ -0,0 +1,8 @@
+PKGNAME=python-is-python3
+PKGSEC=python
+PKGDEP="python-3"
+PKGDES="Symlinks /usr/bin/python to python3"
+PKGPROV="python3"
```
MingcongBai: This is definitely wrong, `python-3` provides `python3` and is not even a virtual package.

MingcongBai (changes requested):

app-games/blastem/autobuild/build
CyberNeko-AI: Corrected.

lang-python/python-is-python3/autobuild/defines
CyberNeko-AI: dropped PKGPROV line.

lang-python/python-2/autobuild/prepare
CyberNeko-AI: Added comment with related info and buid error log.

CyberNeko-AI: I've confirmed that this flag is still needed.

MingcongBai:
@aosc-buildit-bot dickens

MingcongBai:
@aosc-buildit-bot dickens

MingcongBai (approved):
LGTM:

Image

PR #15038, opened 2026-03-05
workflows: retire buildkit-daily.yml

MingcongBai:
We do not appear to be using this workflow and there is no worker receiving this build task. Drop it.


MingcongBai:
Superseded.

PR #15063, opened 2026-03-06
loonggl: only create the egl vendor symlink for the GPU model in use

xry111:
Topic Description
-----------------

We found if both 40_gsgpu.json and 40_loonggpu.json exist, on LG200 PPSSPP will segfault in libEGL_gsgpu.so.  This is need to fix the issue.

Package(s) Affected
-------------------

- loonggl: `1:1.0.2+0ky11.1~0.10-1`

Security Update?
----------------

No

Build Order
-----------

```
#buildit loonggl
```

Test Build(s) Done
------------------

**Primary Architectures**


MingcongBai (approved):
Tested good on OrangePi Nova.

PR #15131, opened 2026-03-11
unixbench: disable lto to prevent dhry Run_Index from being optimized away

AirFortressIlikara:
Topic Description
-----------------

- unixbench: disable lto to prevent dhry Run_Index from being optimized away

Package(s) Affected
-------------------

- unixbench: 6.0.0-2

Security Update?
----------------

No

Build Order
-----------

```
#buildit unixbench
```

Test Build(s) Done
------------------

**Primary Architectures**

- [ ] AMD64 `amd64`
- [ ] AArch64 `arm64`

**Secondary Architectures**

- [ ] Loongson 3 `loongson3`
- [ ] RISC-V 64-bit `riscv64`

### DIFF
--- a/app-benchmarks/unixbench/autobuild/build
+++ b/app-benchmarks/unixbench/autobuild/build
@@ -1,4 +1,4 @@
-make -C UnixBench/ OPTON="${CPPFLAGS} ${CFLAGS} -std=gnu17 ${LDFLAGS}"
+make -C UnixBench/ OPTON="${CPPFLAGS} ${CFLAGS} ${LDFLAGS}"
 
 mkdir -p "$PKGDIR"/usr/{bin,lib/unixbench}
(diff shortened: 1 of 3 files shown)


app-benchmarks/unixbench/autobuild/build
```diff
@@ -1,4 +1,4 @@
-make -C UnixBench/ OPTON="${CPPFLAGS} ${CFLAGS} -std=gnu17 ${LDFLAGS}"
+make -C UnixBench/ OPTON="${CPPFLAGS} ${CFLAGS} ${LDFLAGS}"
```
xry111, line 1: Why change this?  "Legacy" code really requires downgrading the standard.  If even gnu17 does not work try gnu99 or even gnu89.

xry111 (changes requested):
See also https://gcc.gnu.org/PR117695 but note that the report is invalid.

PR #15415, opened 2026-04-01
linux-pam: fix container login failure

ImBearChild:
Topic Description
-----------------

- Remove the restrictive default rules (* - rtprio 0 and * - nice 0) that were causing login failures in containers (systemd-nspawn, LXC, Docker, Podman, etc.). These containers lack the CAP_SYS_NICE capability neededto set nice values, causing pam_limits.so to fail authentication when marked as 'required' in /etc/pam.d/system-auth.

- Move the audio group limits to overrides/etc/security/limits.d/ instead of patching limits.conf.

Package(s) Affected
-------------------

- linux-pam 1.7.1

Security Update?
----------------

No

Build Order
-----------

```
#buildit linux-pam 
```

Test Build(s) Done
------------------

**Primary Architectures**

- [x] AMD64 `amd64`
- [ ] AArch64 `arm64`

**Secondary Architectures**

- [ ] Loongson 3 `loongson3`
- [ ] RISC-V 64-bit `riscv64`


xry111:
It's not a good idea to impact the desktop system's security for container.  That deviates from the [design purpose](https://aosc.io/aosc-os/right-for-me) of the entire project.

ImBearChild:
> impact the desktop system's security

With respect, there’s a misunderstanding.

This patch is not about security, despite files being placed under `/etc/security`. The removed `rtprio` and `nice` lines relate to process priorities, not access control. The relocated `@audio` entries are intended to empower realtime audio services, which also aren’t security settings.

Major distributions such as Fedora, Debian, and NixOS do not include this extra configuration and run fine in containers. We should follow established practice, since there no good reason to go against it.

xry111:
OK.  I guess RTKit is a cleaner solution for real-time audio apps.

app-admin/linux-pam/spec
```diff
@@ -2,3 +2,4 @@ VER=1.7.1
 SRCS="tbl::https://github.com/linux-pam/linux-pam/releases/download/v$VER/Linux-PAM-$VER.tar.xz"
 CHKSUMS="sha256::21dbcec6e01dd578f14789eac9024a18941e6f2702a05cf91b28c232eeb26ab0"
 CHKUPDATE="anitya::id=12244"
+REL=2
```
xry111:
```suggestion
REL=1
```

xry111 (changes requested):
REL should start from 1 (i.e. w/o REL means REL=0).

ImBearChild:
> OK. I guess RTKit is a cleaner solution for real-time audio apps.

PipeWire will fall back to the RTKit D-Bus service when a user cannot obtain real-time priorities via rlimits. Those rules are therefore not strictly necessary in typical setups, but I recommend keeping them for compatibility: they likely exist for a reason we may not know, and removing them could break some users' configurations.

app-admin/linux-pam/spec
ImBearChild: Okay. I will fix this.

xry111 (approved):

PR #15418, opened 2026-04-02
PULL_REQUEST_TEMPLATE: drop the Security Update section

xry111:
Apparently nobody wants to care this now as we already have TUM.


MingcongBai:
See https://github.com/AOSC-Dev/aosc-os-abbs/pull/12680. Coordinated updates needed on multiple projects.

PR #16190, opened 2026-06-03
gcc-14: update to 14.3.0-2

xry111:
Topic Description
-----------------

gcc-14: update to 14.3.0-2

- Backport 5 patches from gcc-15 as they fix bugs affecting 14.3.0 as well.
  - This allows to get rid of the ad-hoc handling of $TRIPLE-gcc-tmp.
- Add a hunk missed in the text domain patch (persumably lefted when we renamed gcc to gcc-14 at the end of Core 12), so gcc-14 can find the translations.
- Remove --with-isl=/usr like gcc-16, to avoid issues like linking gcc-14 libstdc++.so against AOSC Core gcc-15 libgcc_s.so.
  - In the future (Core 14?) we should stop providing (unversioned) /usr/lib/libgcc_s.so and /usr/lib/libgcc*.a.  No other distros supporting multi GCC version (i.e. using --enable-version-specific-runtime-libs) do this, AFAIK.
- Tracking AOSC patches at AOSC-Tracking/gcc @ aosc/releases/gcc-14.3.0 (HEAD: eff96f44bcf49555b24b3b2e32e4d08f1fb4ee33).

Package(s) Affected
-------------------

- gcc-14: `14.3.0-2`
- gcc-runtime-14: `14.3.0-2`

Security Update?
----------------

No

Build Order
-----------

```
#buildit gcc-14
```

Test Build(s) Done
------------------

**Primary Architectures**

- [x] AMD64 `amd64`
- [x] AArch64 `arm64`

**Secondary Architectures**

- [x] Loongson 3 `loongson3`
- [x] RISC-V 64-bit `riscv64`

### DIFF
--- a/app-devel/gcc-14/01-runtime/beyond
+++ b/app-devel/gcc-14/01-runtime/beyond
@@ -113,4 +113,4 @@ fi
 
 abinfo "Installing libgo symlinks ..."
 LIBGO="$(find "$PKGDIR"/"$GCC_LIBDIR" -type l -name 'libgo.so.*' -printf '%f\n')"
-ln -svf "${GCC_LIBDIR##/usr/lib/}"/"$LIBGO".0.0 "$PKGDIR"/usr/lib/"$LIBGO"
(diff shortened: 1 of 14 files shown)


xry111:
截图 2026-06-05 08-16-05

The message translation is fixed.

MingcongBai (approved):

PR #16495, opened 2026-06-24
samba: drop the hand-brew systemd units

xry111:
Topic Description
-----------------

The systemd units shipped in the Samba source package are already installed,
and they seem fit better into the systemd enviroment (with Type=notify
instead of forking in the hand-brew units).  More seriously, the
smbd.socket/smbd@.service combination is completely broken and we have
have no way to fix it without some pheasant patch (likely non-exist as
at now) adding systemd socket activation support into the Samba code
base.

- aosc-os-presets-desktop: Enable smb.service instead of smbd.socket
- samba: drop the hand-brew systemd units

Package(s) Affected
-------------------

- aosc-os-presets-desktop: `9`
- samba: `4.24.3-1`
- ldb: `4.24.3-1`

Security Update?
----------------

No

Build Order
-----------

```
#buildit aosc-os-presets-desktop samba
```

Test Build(s) Done
------------------

**Primary Architectures**

- [x] AMD64 `amd64`
- [x] AArch64 `arm64`

**Secondary Architectures**

- [x] Loongson 3 `loongson3`
- [x] RISC-V 64-bit `riscv64`


app-network/samba/01-samba/postinst
```diff
@@ -1,2 +1,48 @@
...
+
+if [ -d /run/systemd/system ]; then
+  echo "Restarting the running Samba service units (if any) ..."
+  systemctl daemon-reload
+  systemctl try-restart {smb,nmb,winbind,ctdb,samba,samba-bgqd}.service
+fi
+
+if [ -n "${to_be_enable}" ]; then
+  echo "Migrating from old Samba build: enabling ${to_be_enable} ..."
+  systemctl enable ${to_be_enable}
```
MingcongBai: Should we use `systemctl preset` here to enable, and possibly add `--now` so that it takes effect immediately (this was done in `try-restart` but it seemed fragmented)?

app-network/samba/01-samba/postinst
```diff
@@ -1,2 +1,48 @@
...
 systemd-tmpfiles --create samba.conf
 ldconfig
+
+if [ -d /run/systemd/system ]; then
+  echo "Restarting the running Samba service units (if any) ..."
+  systemctl daemon-reload
+  systemctl try-restart {smb,nmb,winbind,ctdb,samba,samba-bgqd}.service
+fi
+
+if [ -n "${to_be_enable}" ]; then
```
MingcongBai: to_enable, or to_be_enabled

app-network/samba/01-samba/postinst
```diff
@@ -1,2 +1,48 @@
...
+  echo "Migrating from old Samba build: enabling ${to_be_enable} ..."
+  systemctl enable ${to_be_enable}
+fi
+
+if [ -n "${to_be_start}" ]; then
+  echo "Migrating from old Samba build: starting ${to_be_start} ..."
+  systemctl start ${to_be_start}
```
MingcongBai: Seems like all three routines could be somewhat combined

MingcongBai (changes requested):

app-network/samba/01-samba/postinst
xry111: I'd keep smb.service disabled if the user has disabled all of smd.service, and (the removed) smbd.service and smbd.socket.  But `preset` will always enable it with new aosc-os-presets-desktop, or do nothing with old aosc-os-presets-desktop.

Also there seems no obvious way to introduce a dependency of samba on aosc-os-presets-desktop, so we cannot assume the version of aosc-os-presets-desktop here.  We also want to handle the case for e.g. someone is running a NAS based on AOSC Base (instead of AOSC Desktop) and smbd.service has been enabled manually, so relying on presets-desktop is not a good idea.

IMO "enable" and "start" are orthogonal concepts (enable --now is more like a grammar sugar instead of a solid concept).  A service can be disabled but running, or enabled but not running.  So for a normal upgrade, we should only do try-restart (i.e. the "enable" status and "start" status will be both preserved, but the already started services are restarted so the old binary will not be stale in the memory).  For example, if someone upgrades Samba from 4.24.3-1 (this version) to 4.24.4 (future version), try-restart will be sufficient (to_be_enable and to_be_start will be empty then).

But for the 4.24.3 => 4.24.3-1 upgrade, we need to migrate the status of smbd.{socket,service} to smb.service, and winbindd.service to winbind.service, etc.  So the two loops exist.

app-network/samba/01-samba/postinst
xry111: Done.

app-network/samba/01-samba/postinst
xry111: Based on the previous comment IMO they are for different (though perhaps related) purposes.

app-network/samba/01-samba/postinst
MingcongBai: Fair enough.

xry111:
截图 2026-06-25 20-41-42

Smoke tested in an amd64 VM.

MingcongBai (approved):

PR #16665, opened 2026-07-08
taisei: update to 1.4.5

LiarOnce:
Topic Description
-----------------

 - cglm: new, 0.9.6
 - taisei: update to 1.4.5

Starting with version 1.4.x, taisei has been upgraded to SDL3; it now requires `cglm` and no longer requires dependencies such as `libzip`.

Package(s) Affected
-------------------

 - cglm: new, 0.9.6
 - taisei: update to 1.4.5

Security Update?
----------------

No

Build Order
-----------

```
#buildit cglm taisei
```

Test Build(s) Done
------------------

**Primary Architectures**

- [x] AMD64 `amd64`
- [x] AArch64 `arm64`

- [ ] 32-bit Optional Environment `optenv32`

**Secondary Architectures**

- [x] Loongson 3 `loongson3`
- [x] RISC-V 64-bit `riscv64`


app-scientific/cglm/autobuild/defines
```diff
@@ -0,0 +1,7 @@
+PKGNAME=cglm
+PKGSEC=math
+BUILDDEP="meson ninja"
+PKGDES="Highly Optimized 2D / 3D Graphics Math (glm) for C"
+PKGPROV="libcglm0 libcglm-doc libcglm-dev"
```
MingcongBai: This should have been automatically handled by Spiral.

app-scientific/cglm/autobuild/defines
```diff
@@ -0,0 +1,7 @@
+PKGNAME=cglm
+PKGSEC=math
+BUILDDEP="meson ninja"
+PKGDES="Highly Optimized 2D / 3D Graphics Math (glm) for C"
```
MingcongBai: Optmised 2D/3D OpenGL Mathematics (GLM) library

app-scientific/cglm/autobuild/defines
```diff
@@ -0,0 +1,7 @@
+PKGNAME=cglm
+PKGSEC=math
+BUILDDEP="meson ninja"
```
MingcongBai, line 4: No ELF dependency at all?

app-scientific/cglm/spec
```diff
@@ -0,0 +1,5 @@
+VER=0.9.6
+REL=1
```
MingcongBai: Drop this line as it's new.

app-games/taisei/autobuild/defines
```diff
@@ -1,8 +1,8 @@
 PKGNAME=taisei
 PKGSEC=games
-PKGDEP="glu hicolor-icon-theme libpng mesa sdl2-image sdl2-mixer sdl2-ttf \
-        freealut libzip"
-BUILDDEP="meson ninja docutils"
+PKGDEP="mesa sdl3 sdl3-image sdl3-ttf cglm libunibreak libwebp zstd freetype opusfile zlib"
+BUILDDEP="gcc meson ninja gettext python-3"
+PKGSUG="gamemode glslang libpng"
```
MingcongBai: I don't believe that glslang and libpng could have been Suggests, maybe for gamemode (what does it do?).

MingcongBai (changes requested):

app-games/taisei/autobuild/defines
Akaricchi: glslang and libpng are optional dependencies but they are linked, not conditionally loaded at runtime. So if the game is built with them, then they are required at runtime. [Gamemode](https://github.com/feralinteractive/gamemode) support is detected at runtime; the game will work fine if it's not installed.

glslang is only needed for OpenGL ES and SDL_GPU rendering backends. Safe to skip if you don't need them, required otherwise. It's also linked in.

Only the headers are used from cglm, so technically it's a build-time-only dependency.

MingcongBai (approved):

PR #16775, opened 2026-07-15
linux-kernel{,-rc}: drop loongson.ls7a[12]000_support patches

xry111:
Topic Description
-----------------

We can no longer reproduce the boot failures with 7A1000 even with the
loongson.ls7a1000_support=1 opt-in.
    
For 7A2000, we've shipped a softdep in loonggpu-kernel-dkms so loonggpu
is always loaded before drm/loongson.  Thus we no longer need to disable
the 7A2000 support to avoid a conflict with loonggpu.
    
To opt-out drm/loongson, we can use modprobe.blacklist=loongson anyway
so we don't need to keep the custom options just for opt-out.
    
So drop the patches adding custom options.

Package(s) Affected
-------------------

- linux-kernel-7.0.13: `7.0.13-3`
- linux-kernel-rc-7.1.0: `7.1.0-0.7a`

Security Update?
----------------

No

Build Order
-----------

```
#buildit linux-kernel linux-kernel-rc
```

Test Build(s) Done
------------------

**Primary Architectures**

- [ ] AMD64 `amd64`
- [ ] AArch64 `arm64`

**Secondary Architectures**

- [ ] Loongson 3 `loongson3`
- [ ] RISC-V 64-bit `riscv64`


MingcongBai:
Will handle in the next mainline and RC kernel update. Coming up for testing today.

PR #17248, opened 2026-08-20
rtl-sdr: update to 2.0.3

LengmoAngel:
## Topic Description

Update rtl-sdr from 2.0.2 to 2.0.3.

## Package(s) Affected

rtl-sdr

## Security Update?

No

## Build Order

```text
#buildit rtl-sdr
```

## Test Build(s) Done

**Primary Architectures**

* [ ] AMD64 `amd64`
* [ ] AArch64 `arm64`
* [ ] LoongArch 64-bit `loongarch64`
* [ ] LoongArch 64-bit (No SIMD) `loongarch64_nosimd`

**Secondary Architectures**

* [ ] Loongson 3 `loongson3`
* [ ] PowerPC 64-bit (Little Endian) `ppc64el`
* [ ] RISC-V 64-bit `riscv64`

### DIFF
--- a/app-devices/rtl-sdr/autobuild/defines
+++ b/app-devices/rtl-sdr/autobuild/defines
@@ -3,3 +3,4 @@ PKGDES="Driver for Realtek RTL2832U that enables general-purpose, software-defin
 PKGDEP="glibc libusb"
 BUILDDEP="cmake"
 PKGSEC="hamradio"
+ABTYPE=cmakeninja
--- a/app-devices/rtl-sdr/spec
+++ b/app-devices/rtl-sdr/spec
@@ -1,4 +1,4 @@
-VER=2.0.2
+VER=2.0.3
 SRCS="git::commit=tags/v${VER}::https://gitea.osmocom.org/sdr/rtl-sdr.git"
 CHKSUMS="SKIP"
 CHKUPDATE="anitya::id=17786"


app-devices/rtl-sdr/spec
stdmnpkg, line 1: in autobuild/defines:

specify ABTYPE

LengmoAngel: Done, added ABTYPE=cmake. Thanks!

app-devices/rtl-sdr/autobuild/defines
```diff
@@ -3,3 +3,4 @@ PKGDES="Driver for Realtek RTL2832U that enables general-purpose, software-defin
 PKGDEP="glibc libusb"
 BUILDDEP="cmake"
 PKGSEC="hamradio"
+ABTYPE=cmake
```
stdmnpkg: try cmakeninja first, use cmake only when it does not works

LengmoAngel: Switched to cmakeninja and verified that it builds successfully with Ciel on amd64.

OriginCode:
You may attach your test result for the package (a basic functional test would work). Otherwise LGTM.

MingcongBai:
> You may attach your test result for the package (a basic functional test would work). Otherwise LGTM.

@LengmoAngel ping

LengmoAngel:
Basic functional test done on `amd64` without RTL-SDR hardware attached.

Tested the clean `rtl-sdr` 2.0.3 package built with Ciel (`ABTYPE=cmakeninja`).

`rtl_test -h` works as expected:

```text
rtl_test, a benchmark tool for RTL2832 based DVB-T receivers

Usage:
    [-s samplerate (default: 2048000 Hz)]
    [-d device_index or serial (default: 0)]
    [-t enable Elonics E4000 tuner benchmark]
    [-p[seconds] enable PPM error measurement (default: 10 seconds)]
    [-b output_block_size (default: 16 * 16384)]
    [-S force sync output (default: async)]
```

All packaged `rtl_*` binaries load the packaged `librtlsdr.so.0` successfully. For example:

```text
=== rtl_test ===
librtlsdr.so.0 => /tmp/rtl-sdr-2.0.3-test/usr/lib/librtlsdr.so.0
```

No missing dynamic libraries were found.

`pkg-config` reports the expected version:

```text
$ pkg-config --modversion librtlsdr
2.0.3
```

I also built a small C smoke test against the packaged `rtl-sdr.h` and `librtlsdr.so` and called `rtlsdr_get_device_count()`:

```text
librtlsdr API call OK
device_count=0
```

`device_count=0` is expected since no RTL-SDR device is attached.

Also verified the packaged auxiliary files:

```text
udev rule: PASS
modprobe config: PASS
header: PASS
pkg-config file: PASS
```

Hardware-dependent functionality was not tested due to the lack of an RTL-SDR device.

MingcongBai (approved):

PR #17372, opened 2026-08-26
libdrm-loonggpu: new, 2.4.134

Red54:
Topic Description
-----------------

libdrm-loonggpu, named in reference to libdrm-gsgpu (the predecessor of libldrm), provides libldrm.so.1 via update-alternatives.

References:
https://lauosc.cn/d/303-gsgpu/18
https://github.com/AOSC-Dev/aosc-os-abbs/pull/17175

Package(s) Affected
-------------------

libdrm-loonggpu

Security Update?
----------------

No

Build Order
-----------

```
#buildit libdrm-loonggpu
```

Test Build(s) Done
------------------

**Primary Architectures**

- [ ] AMD64 `amd64`
- [ ] AArch64 `arm64`

**Secondary Architectures**

- [ ] Loongson 3 `loongson3`
- [ ] RISC-V 64-bit `riscv64`

### DIFF
--- a/runtime-display/ldrm/autobuild/build
+++ b/runtime-display/ldrm/autobuild/build
@@ -5,7 +5,7 @@ dpkg -x "$SRCDIR"/libldrm_${__UPSTREAM_VER}_loong64.deb \
 abinfo "Moving libraries to the correct location ..."
 mv -v "$PKGDIR"/usr/lib/loongarch64-linux-gnu/* \
     "$PKGDIR"/usr/lib/
-rm -rv "$PKGDIR"/usr/lib/loongarch64-linux-gnu
+rm -rv "$PKGDIR"/usr/lib/{loongarch64-linux-gnu,libldrm.so.1}
 
 abinfo "Setting executable bits for shared objects ..."
 chmod -v +x "$PKGDIR"/usr/lib/*.so*
--- a/runtime-display/libdrm-loonggpu/autobuild/build
+++ b/runtime-display/libdrm-loonggpu/autobuild/build
@@ -1,0 +1,16 @@
+SO_NAME=libdrm_loonggpu.so.1.${PKGVER##*.}.0
+
+build_meson_configure
+
+abinfo "Building binaries ..."
+ninja -C $BLDDIR amdgpu/$SO_NAME tests/amdgpu/loonggpu_stress
+
+abinfo "Installing binaries ..."
+mkdir -pv $PKGDIR/usr/lib
+cp -v $BLDDIR/amdgpu/$SO_NAME $PKGDIR/usr/lib
+ln -sv $SO_NAME $PKGDIR/usr/lib/libdrm_loonggpu.so.1
+ln -sv libdrm_loonggpu.so.1 $PKGDIR/usr/lib/libdrm_loonggpu.so
+mkdir -pv $PKGDIR/usr/bin
+cp -v $BLDDIR/tests/amdgpu/loonggpu_stress $PKGDIR/usr/bin
+mkdir -pv $PKGDIR/usr/share/libdrm
+cp -v data/loonggpu.ids $PKGDIR/usr/share/libdrm
--- a/runtime-display/ldrm/autobuild/alternatives
+++ b/runtime-display/ldrm/autobuild/alternatives
@@ -1,0 +1,1 @@
+alternative /usr/lib/libldrm.so.1 /usr/lib/libldrm.so.1.0.0 40
--- a/runtime-display/ldrm/spec
+++ b/runtime-display/ldrm/spec
@@ -7,3 +7,4 @@ SRCS="file::use-url-name=true::https://repo.aosc.io/aosc-repacks/loonggpu/libldr
 CHKSUMS="sha256::bc1ba9cd584a339a755926628ca42080ca508b7aa571a1ef885074d4fdb90bde"
 SUBDIR=.
 CHKUPDATE="anitya::id=376483"
+REL=1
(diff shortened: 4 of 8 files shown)


Icenowy:
I think in modern days new GPU drm interface is placed directly into Mesa instead of making new subsidiary library in libdrm, and libraries in libdrm are being removed (e.g. freedreno is removed in libdrm master).

xry111:
> I think in modern days new GPU drm interface is placed directly into Mesa instead of making new subsidiary library in libdrm, and libraries in libdrm are being removed (e.g. freedreno is removed in libdrm master).

This is intended to be a hack against the stupid vendor driver stack instead of a part of a really modern driver.

runtime-display/libdrm-loonggpu/autobuild/beyond
xry111, line 1: Use the autobuild4 mechanism for alternatives instead of home-brewing it.

Also the libldrm package will need to be changed to use the alternatives too, if we really need the alternatives: IMO it may be easier to just replace the libldrm package with this, in the test phase we can simply use oma topic and after the test phase I don't think we really want to keep two implementations.

Red54: Done, but I still kept ldrm, just in case big changes later.

xry111: Then it will be very difficult to allow 4-KiB configuration in dkms as there's no way for the kernel module to know which alternative is selected.

runtime-display/libdrm-loonggpu/autobuild/build
```diff
@@ -0,0 +1,16 @@
...
+abinfo "Building binaries ..."
+ninja -C $BLDDIR amdgpu/$SO_NAME tests/amdgpu/loonggpu_stress
+
+abinfo "Installing binaries ..."
+mkdir -pv $PKGDIR/usr/lib
+cp -v $BLDDIR/amdgpu/$SO_NAME $PKGDIR/usr/lib
+ln -sv $SO_NAME $PKGDIR/usr/lib/libdrm_loonggpu.so.1
+ln -sv libdrm_loonggpu.so.1 $PKGDIR/usr/lib/libdrm_loonggpu.so
+mkdir -pv $PKGDIR/usr/bin
+cp -v $BLDDIR/tests/amdgpu/loonggpu_stress $PKGDIR/usr/bin
```
xry111, line 14: The convention of AOSC OS is not to install test programs.

runtime-display/ldrm/autobuild/build
```diff
@@ -5,7 +5,7 @@ dpkg -x "$SRCDIR"/libldrm_${__UPSTREAM_VER}_loong64.deb \
 abinfo "Moving libraries to the correct location ..."
 mv -v "$PKGDIR"/usr/lib/loongarch64-linux-gnu/* \
     "$PKGDIR"/usr/lib/
-rm -rv "$PKGDIR"/usr/lib/loongarch64-linux-gnu
+rm -rv "$PKGDIR"/usr/lib/{loongarch64-linux-gnu,libldrm.so.1}
```
xry111, line 8: Keeping /usr/lib/libldrm.so.1.0.0 makes ldconfig to recreate the libldrm.so.1 symlink against it, even if the symlink currently points to /etc/alternatives/libldrm.so.1.  You need to move the shared libraries out of /usr/lib to prevent ldconfig from doing that.

Note that ldconfig is frequently ran by the maintainer scripts of all the packages so it's not a theoretical only issue.

runtime-display/libdrm-loonggpu/autobuild/build
```diff
@@ -0,0 +1,16 @@
+SO_NAME=libdrm_loonggpu.so.1.${PKGVER##*.}.0
+
+build_meson_configure
+
+abinfo "Building binaries ..."
+ninja -C $BLDDIR amdgpu/$SO_NAME tests/amdgpu/loonggpu_stress
+
+abinfo "Installing binaries ..."
+mkdir -pv $PKGDIR/usr/lib
+cp -v $BLDDIR/amdgpu/$SO_NAME $PKGDIR/usr/lib
```
xry111, line 10: The shared library has a broken DT_RUNPATH ($ORIGIN/.., which is /usr and it does not make sense at all) because it's supposed to be used in the build tree.  Perform a DESTDIR install first and then copy it from DESTDIR to fix the issue.

runtime-display/libdrm-loonggpu/autobuild/beyond
xry111: The more I think the more I believe the old ldrm package should just go away.  We have Git and PKGEPOCH so it's really trivial to revert to ldrm in case doing so is really necessary.

runtime-display/ldrm/autobuild/build
xry111: 截图 2026-09-06 05-45-03

xry111 (changes requested):

runtime-display/libdrm-loonggpu/autobuild/build
Red54: However, `libdrm` installs test programs.
```
$ dpkg -L libdrm |grep bin
/usr/bin
/usr/bin/amdgpu_stress
/usr/bin/drmdevice
/usr/bin/etnaviv_2d_test
/usr/bin/etnaviv_bo_cache_test
/usr/bin/etnaviv_cmd_stream_test
/usr/bin/modeprint
/usr/bin/modetest
/usr/bin/proptest
/usr/bin/vbltest
```
Also, loonggpu_stress can test GPU memory allocation and read/write operations. I'm inclined to keep it, what do you think?

PR #17663, opened 2026-09-14
micro: fix missing of syntax highlighting

Suyun114:
Topic Description
-----------------

- micro: fix missing of syntax highlighting
    Syntax highlighting was missing after switching from \`ABTYPE=self\` to
    \`ABTYPE=gomod\`, because of \`ABTYPE=gomod\` would not run \`go generate ...\`
    to generate syntax file headers. So we need to generate them manually
    before building, to provide syntax file headers for syntax highlighting.
    This commit is authored with assistance from AI/LLM:
    - Model: deepseek-v4.1-flash
    - Platform: DeepSeek API \(api.deepseek.com\)
    - Agent platform: Codex \(github.com/openai/codex\)
    - Prompt:
    如此打包的 micro 没有语法高亮，为什么？

Package(s) Affected
-------------------

- micro: 2.0.15-6

Security Update?
----------------

No

Build Order
-----------

```
#buildit micro
```

Test Build(s) Done
------------------

**Primary Architectures**

- [x] AMD64 `amd64`
- [x] AArch64 `arm64`

**Secondary Architectures**

- [x] Loongson 3 `loongson3`
- [x] RISC-V 64-bit `riscv64`


Suyun114:
Tested on amd64.

图片

MingcongBai:
@aosc-buildit-bot dickens

MingcongBai (dismissed):

app-editors/micro/autobuild/prepare
```diff
@@ -4,3 +4,6 @@ GO_LDFLAGS=(
     "-X 'github.com/zyedidia/micro/v2/internal/util.CommitHash=$(git rev-parse --short HEAD)'"
     "-X 'github.com/zyedidia/micro/v2/internal/util.CompileDate=$(date +%Y-%m-%d)'"
```
scratch-er, line 5: I think you have better add a comment above the GO_LDFLAGS here, explaining what they are doing, since this flags are package specific and their intentions are not obvious. Although these GO_FLAGS are not written by you, adding such a comment would be a good packaging style.

scratch-er:
I also wonder whether "`go generate` is needed for syntax highlighting" is documented in upstream docs. Seems that upstream is using a Makefile to automate the entire process. However:

- If building without features like syntax highlighting is supported upstream, it should be documented.
- If this is not supposed to be supported upstream, building without running `go generate` should have caused an error, rather silently produce binaries with missing features.

If the upstream does have such problem on build system or documentation, it is worth reporting to upstream.

app-editors/micro/autobuild/prepare
MingcongBai: I think this was already addressed in the commit message.

As for upstream routines - this wasn't a common practice anyway, so it could be worth bringing up to the upstream. @Suyun114 

Approving as-is.

MingcongBai (approved):

